PR #10, opened 2022-10-08
Rewrite main menu positioning system, round text to integer pixels

nyanpasu64:
Previously, main menu text would show diagonal seams on some GPUs and window sizes. This occurred when menu text was positioned at a half-pixel horizontal/vertical offset, causing pixel boundaries to lie in the middle of screen pixels.

The new code rounds menu item positions to integer pixel coordinates. Along with 228f1c9 (which places menu text at integer coordinates relative to menu items), this fixes diagonal seams on main menu text.

It also generalizes better to more than 4 menu items.

- [ ] I inserted a lot of asserts for debugging, and to verify that `_item_position(offset)` had integer coordinates (currently the case, future changes may invalidate this assumption and trip the assert). IDK if it slows down the game and should be removed.
- [ ] I didn't add or omit type hints consistently, `_item_position(offset)` lacks a hint.
- [ ] Was adding `_clamp_cycle_step` the right move? I think it's a good thing since it prevents theoretical numeric overflow if a user continually scrolls the same direction indefinitely. Now you can probably remove various `posmod` when *checking* `cycle_step`, but I didn't try to accomplish this.
- [ ] Should I omit leading underscores for non-override methods?
- [ ] The comment in `step()` is accurate to the best of my knowledge. Are there any errors, should I revise it further for clarity, move it elsewhere in the code, remove the "Why." comment, or (in the future) rewrite the underlying code so it doesn't list menu items twice in opposite directions for drawing and selecting?

Any issues I missed?

### DIFF
--- a/scenes/menus/title/main_menu/main_menu.gd
+++ b/scenes/menus/title/main_menu/main_menu.gd
@@ -33,7 +33,8 @@ var num_items : int  # Modulo for cycle_step.
 var show_options = false
 
 
-func _clamp_cycle_step():
+func _cycle_increment(cycle_direction: int) -> void:
+	cycle_step += cycle_direction
 	cycle_step = posmod(cycle_step, num_items)
 
 
@@ -82,14 +83,14 @@ func _process(delta):
 			center_pos_idx = 2
 			
 			var items = [[story, selector_story], [settings, selector_settings], [extra, selector_extra], [ld, selector_ld]]
-			# Be sure to initialize num_items before step() -> _clamp_cycle_step() reads it.
+			# Be sure to initialize num_items before _cycle_through() -> _clamp_cycle_step() reads it.
 			num_items = items.size()
 			
 			if Input.is_action_just_pressed("left"):
-				step(-1)
+				_cycle_through(-1)
 			
 			if Input.is_action_just_pressed("right"):
-				step(1)
+				_cycle_through(1)
 			
 			var item_progress = sin(cycle_progress * PI/2)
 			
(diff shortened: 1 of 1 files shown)


nyanpasu64:
poke

GTcreyon:
> poke

I'll get to this soon, promise. :)

GTcreyon:
Alright, let's have a look. ^^

- Asserts don't slow down the game, don't worry. They're also not included in release builds. Adding them is good practice.
- I'll try to add some type hints before merging, but it's not too important.
- Yes, `_clamp_cycle_step()` is good, but I think I'll replace it with a `_increment_cycle_step()` in order to include both the increment/decrement and also the clamping in one function.
- No, underscores are good. They're used to denote *private* methods, which all of the ones you used are.
- Honestly, I don't remember how `step()` works. You seem to know what you're doing, so I'll assume it's correct. I'll rename it to `_cycle_through()` though, since it's marginally clearer than "step".

Thank you for the very thorough PR!

scenes/menus/title/main_menu/main_menu.gd
```diff
@@ -82,14 +83,14 @@ func _process(delta):
 			center_pos_idx = 2
 			
 			var items = [[story, selector_story], [settings, selector_settings], [extra, selector_extra], [ld, selector_ld]]
-			# Be sure to initialize num_items before step() -> _clamp_cycle_step() reads it.
+			# Be sure to initialize num_items before _cycle_through() -> _clamp_cycle_step() reads it.
```
nyanpasu64, line 86: Change comment to say `_cycle_through() -> _cycle_increment()`.

GTcreyon:
Alright, I'm happy to merge now.

scenes/menus/title/main_menu/main_menu.gd
GTcreyon: Whoops, didn't see that, I'll fix that on the master branch.

GTcreyon: Nevermind, it's fine. I disliked the name `_cycle_increment()` so I changed it. Apologies, I misread your note.

PR #12, opened 2022-10-21
Fix and improve on warp pipes

Koopa1018:
Previously, when Mario entered a warp pipe, the slide-down animation would get interrupted midway by the following error:
```
Invalid get index 'state' (on base: 'Nil').
```
The game would then softlock with Mario still in the pipe.

This bug occurred because Mario, when sliding, would exit the Area2D responsible for checking if he's on top of the pipe. When Mario exits this area, the pipe nulls its reference to him--but this was happening while he was still in the pipe, making it impossible for the pipe to finish sliding him.

To fix this, the code was altered not to null Mario's reference if he's sliding down the pipe when he leaves the Area2D, and the Area2D itself was extended downwards so Mario doesn't leave it during the slide.

- [ ] In process of making this fix, I rearranged the code to follow the code conventions, cleaned it up some, and commented it for newcomers. You want me to fix or revert any of this?
- [ ] How about the variable renaming? Is that liable to break anything? Cause confusion?

### DIFF
--- a/classes/interactable/pipe/exit_pipe.gd
+++ b/classes/interactable/pipe/exit_pipe.gd
@@ -1,0 +1,8 @@
+extends Pipe
+
+onready var sweep_effect = $"/root/Singleton/WindowWarp"
+
+export var scene_path : String
+
+func _warp(pos):
+	sweep_effect.warp(pos, scene_path)
--- a/classes/interactable/pipe/exit_pipe.gd
+++ b/classes/interactable/pipe/exit_pipe.gd
@@ -1,8 +1,0 @@
-extends Pipe
-
-onready var sweep_effect = $"/root/Singleton/WindowWarp"
-
-export var scene_path : String
-
-func _warp(pos):
-	sweep_effect.warp(pos, scene_path)
--- a/classes/interactable/pipe/pipe.gd
+++ b/classes/interactable/pipe/pipe.gd
@@ -89,9 +89,9 @@ func _on_mario_top(body):
 
 
 func _on_mario_off(_body):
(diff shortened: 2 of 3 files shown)


Koopa1018:
Added an option for a pipe to send Mario to another scene, much like a warp zone or door.

- [ ] As with warp zones, scenes are referenced by path, not as `PackedScene`s. Is this preferred? I've been assuming, since warp zones use paths.
- [ ] Pipes can be switched between position-warp and scene-warp modes using a checkbox. This design was copied from doors. I experimented with using a separate subclass for scene-warp pipes than for position-warp, but decided it would be easier for designers to use this than a whole extra duplicate pipe prefab.
- [ ] The scene change is done using `WindowWarp,` to achieve a star-shaped iris transition rather than a directional wipe. 
- [ ] I set the iris-out timing where I think it looks good. Happy to tinker with it if there are other thoughts.
- [ ] Iris-in timing currently matches WindowWarp's default value (i.e. very fast). Is this speed by design, or just where it happened to land?

GTcreyon:
- The refactoring all looks good, and is very much appreciated.
- Variable renaming looks fine.
- I'm not quite sure on the path vs. packed front, I'm willing to be swayed, but this is fine for now.
- Yeah, the boolean works well. Someday it may be good to make an exit animation, but this is great for the time being.
- WindowWarp should work nicely.
- Timing is erroneous - it should be 25 for the in-time, not 15. I'll fix that.

GTcreyon (approved):
I'm not quite sure why you changed the size of the collision box, but it works.

PR #14, opened 2022-11-03
Use interactables' global position when Mario slides up to them

Koopa1018:
When Mario interacts with a sign (or presumably a Toad), he shifts over to a designated spot relative to the sign (or Toad).

Vexxter on Discord discovered that when signs are children of other nodes, the parent nodes' positions are not taken into account when Mario shifts like this.

This means that if you load in a scene then move its root node, reading a sign in that scene will teleport Mario somewhere crazy (specifically to the global position with the same coordinates as the sign's local position).

Using the interactable's global position takes parent nodes into account, thereby solving the problem.

Just to be on the safe side, also made sure Mario uses global coordinates for this shift animation. He's a child of the scene root, and the scene root is no more guaranteed to be at (0,0) than any other node....

### DIFF
--- a/classes/interactable/interactable_dialog.gd
+++ b/classes/interactable/interactable_dialog.gd
@@ -9,7 +9,7 @@ onready var dialog = $"/root/Main/Player/Camera/GUI/DialogBox"
 
 
 func _interact_with(body):
-	body.sign_x = global_position.x + x_offset * sign(body.position.x - global_position.x)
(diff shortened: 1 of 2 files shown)


GTcreyon:
Seems all good!

GTcreyon:
Thank you very much for your work! I will probably change some things about how InteractableDialog works, having looked at this for a bit.

PR #17, opened 2022-11-10
Update and improve doors

Koopa1018:
Did some work to nicen up doors and make them consistent with the rest of the project:

- For starters, doors now do scene changes with WindowWarp like pipes do, rather than directly calling `get_tree().change_scene_to(next_scene)`. 
- Moved doors' origins to bottom center to facilitate snapping to the ground.
- Doors now render at Z = -1, behind Mario by default, like signs do.
- Improved the enter-door animation somewhat. The doorframe now stays in place as it should, and Mario now moves over and looks in as the door opens. This is definitely a work-in-progress, waiting on an enter-door animation for Mario, but it's a decent placeholder for now.
- Added ability to switch doors' sprites at runtime.
- Made a dedicated class for double doors. Double doors are internally two normal doors tied together, with a utility script to let you treat them as one door.

- [x] Because of WindowWarp, doors use scene paths instead of `PackedScene` now (also more consistent with pipes and warp zones).
- [x] Moving the origin probably breaks levels that use doors. It doesn't look like there are any such levels, but I just wanted to double check.
- [x] Thoughts on this method of reskinning doors? Is this a good way to do it?

### DIFF
--- a/classes/interactable/door/door.gd
+++ b/classes/interactable/door/door.gd
@@ -28,8 +28,8 @@ func _player_offset() -> int:
 	return 0
 
 
-func _player_begin_animation(mario):
-	mario.switch_anim("back")
+func _player_begin_animation(player):
+	player.switch_anim("back")
 
 
 func _door_begin_animation():
--- a/classes/interactable/door/door.gd
+++ b/classes/interactable/door/door.gd
@@ -1,18 +1,18 @@
-tool
 class_name Door
 extends InteractableWarp
 
 const CENTERING_SPEED = 0.25
 
-export var door_graphic : SpriteFrames setget set_door_graphic
+export var door_graphic : SpriteFrames
 
 var player_fade_start_time : int
 var player_fade_rate : float
 var door_close_start_time : int
 
-func set_door_graphic(graphic : SpriteFrames):
-	door_graphic = graphic
+func _ready():
+	_set_sprite_frames(door_graphic)
 	
+	# Load animation timings
 	if door_graphic is DoorSkin:
 		# Resource has animation timings saved in it; load them.
 		player_fade_start_time = door_graphic.player_fade_start_time
@@ -23,10 +23,6 @@ func set_door_graphic(graphic : SpriteFrames):
 		player_fade_start_time = DoorSkin.DEFAULT_PLAYER_FADE_START
 		player_fade_rate = DoorSkin.DEFAULT_PLAYER_FADE_RATE
 		door_close_start_time = DoorSkin.DEFAULT_DOOR_CLOSE_START
-
-
-func _ready():
-	_set_sprite_frames(door_graphic)
 
 
 func _update_animation(_frame: int, _player):
--- a/classes/interactable/door/door.gd
+++ b/classes/interactable/door/door.gd
@@ -2,6 +2,9 @@ class_name Door
 extends InteractableWarp
 
 const CENTERING_SPEED = 0.25
+const ANIM_PLAYER_BEGIN_FADEOUT = 8
+const ANIM_PLAYER_FADE_RATE = 0.1
+const ANIM_DOOR_BEGIN_CLOSE = 32
 
 export var door_graphic : SpriteFrames
 
@@ -15,9 +18,17 @@ func _update_animation(_frame: int, _player):
 	
 	if _frame == 0:
 		# Start door opening animation
-		_door_begin_animation()
+		_door_open()
 		# Start player's enter animation
 		_player_begin_animation(_player)
+	if _frame >= ANIM_PLAYER_BEGIN_FADEOUT:
+		# Fade player down a little bit
+		_player.sprite.modulate.a -= ANIM_PLAYER_FADE_RATE
+	if _frame == ANIM_DOOR_BEGIN_CLOSE:
+		_door_close()
+	if _frame == _animation_length():
+		# Reset player to fully opaque
+		_player.sprite.modulate.a = 1
 
 
 func _animation_length():
@@ -37,8 +48,11 @@ func _player_begin_animation(player):
 	player.read_pos_x = global_position.x + _player_offset()
 
 
-func _door_begin_animation():
+func _door_open():
 	$Sprite.play("opening")
+
+func _door_close():
+	$Sprite.play("closing")
 
 
 func _set_sprite_frames(sprite_frames: SpriteFrames):
--- a/classes/interactable/door/door.gd
+++ b/classes/interactable/door/door.gd
@@ -60,7 +60,6 @@ func _physics_process(_delta):
 
 
 func _on_mario_touch(body):
-	print_debug("Touched door")
 	if body.state == body.S.NEUTRAL:
 		can_warp = true
 		target = body
@@ -68,6 +67,5 @@ func _on_mario_touch(body):
 
 func _on_mario_off(_body):
 	if !entering: # w/o this check, target will get nulled when animation ends
-		print_debug("left door")
 		can_warp = false # Or else he won't
 		target = null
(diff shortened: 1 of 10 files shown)


Koopa1018:
Huh. I've actually got a WarpTrigger base class in the works! Lemme get that added.

Koopa1018:
HOLY dang the Interactable class is so much more elegant than Pipe code.
Anyway, Door now extends `InteractableWarp`, which itself extends `Interactable`. `InteractableWarp` is designed to begin an animation when Mario interacts with it, then perform a warp (either to a new scene or elsewhere in this scene) when the animation finishes.

Double Doors have been rewritten around this. They no longer contain two normal doors--instead, they are themselves a variant of doors with two sprites. In addition, they have a more sophisticated algorithm to pick which of the two doors Mario will go in: whichever one is on the same side of the origin as Mario--with a _slight_ bias towards whichever one he's currently facing.
This algorithm comes out feeling more logical than the collision-box-based selection was. In addition, the hope is that having a bit of bias will make entering double doors while running feel nicer, as Mario will now be less likely to abruptly switch directions on entry. This could be made even smarter in the future by factoring in Mario's current running speed, but I'm concerned that might make it a bit weird for speedrunners.


Noted glitch: doors fail to interact with Mario when they come after him in the scene tree. Not sure how to fix this.

- [ ] Double Door arbitrarily uses the left door for its `Interactable`-inherited `sprite` field. I will confess, I have no idea what this field is for, so I can only hope this is okay.
- [ ] Currently, Mario is manually lerped to the center of doors. It looks like signs center him by setting Mario's `read_pos_x`, but that centering has a different speed (lerp factor). Doors' current lerp factor feels pretty nice, but I'm wondering if we should use `read_pos_x` instead of manual lerping just for consistency.
- [ ] I want to get pipes inheriting from this base class. Can I throw that in here as well, or should I save that for a future PR?

GTcreyon:
- I don't know why the doors don't work due to scene tree position. I'll check that out. Are you confident that that is the cause?
- I think the `sprite` field could have been restricted to the InteractableDialog class, rather than simply Interactable, since it was only used for allowing NPCs to flip to look towards the player. In retrospect, it might have been best to make a separate class for that entirely, but the `can_pivot` bool works well enough to not bother. Either way, don't worry about it; I'll probably fix that later.
- I agree with using `read_pos_x` for consistency, but I also hear your point about the lerp factor. I think the code for it should be rewritten, and ideally moved out of the Player class (although I wouldn't know how to do that cleanly). Also, in retrospect `lerp()` is not a great function to use for repositioning, due to its lack of ways to customise it, along with asymptotic behavior. It may be wise to use a trig function like `cos()` instead, but it's not a big deal. 
- If you're going to do it for pipes aswell, I'd advise making a separate PR first do address some of the issues with the Interactable class that I detailed above (as long as you feel confident doing so), and *then* make another fresh PR for pipes based on those improvements.
- I like your idea for the bias element. I'll have to test that out later.

Koopa1018:
I'll save the pipe improvements for a future PR, then :)

Probably ought to clarify, I was actually trying to implement the suggested Interactable improvements (option 2) with `InteractableWarp`. I did change the name, because it seems like `WarpTrigger` might be too easy to confuse with `WarpZone` (especially since the warp zone object is in a folder called "trigger"). I'll gladly rename it if you'd like.

Sprite field to allow for NPCs to flip...that makes sense.

As for scene ordering making doors not work, all I know is when the scene is ordered with Player _over_ Door, the door does nothing, and when the scene is ordered Player _under Door, it works.

GTcreyon:
Re: tree order, that's strange, I'll check it out later.

Sidenote, best to refer to "the player" instead of "Mario", given that Luigi is planned to be playable.

GTcreyon:
Ahh, it dawned on me. It won't open because the player has process priority when they're higher on the tree, and therefore they can jump before they're able to open the door, and then they're not on the floor to be able to open it.

Koopa1018:
Hmm, good point about Luigi

Koopa1018:
Now sets `read_pos_x` instead of manual lerping, just for consistency.

GTcreyon:
Loving how this is going!

Regarding the scene tree issue, the core of the issue is as follows:
1. You press the up/jump button. Note, these are separate on some control layouts, so don't assume they're always pressed at the same time.
2. `_physics_process()` is run in scene tree order.
3. If the player is higher on the screen tree, its `_physics_process()` will run before the door's `_physics_process()`.
4. As such, the player jumps.
5. The door then checks if the player is able to open it. The player just jumped, so `body.is_on_floor()` is `false`. So no, no entry occurs.

I believe you can fix this simply by changing the door's Process Priority to `-1`.

Also, the doors currently don't close after going through them. It would be nice to implement that, and maybe the player's opacity or lightness fading out.

Just to note, node names should always be in `PascalCase` (no spaces) rather than `Title Case`. I'm aware there are already a few places where this mistake was made before, not your fault :P

Koopa1018:
Hmm, opacity fading. I was assuming there would be a sprited walk-into-door animation, but this seems like quite a workable animation in its own right!

Ah, whoops! Missed that part of the style guide XP

Koopa1018:
Arright, animation has been fully implemented.

Also added a new type, `DoorSkin`, which can be attached to a `SpriteFrames` resource to adjust the timing of the scripted animation. If the resource doesn't have this script, the animation will use default timings.

- [ ] This method of adjusting timings (make sure the SpriteFrames resource has a specific script) seems like it would be hard for new devs to intuit. Should this method be scrapped because of that? If not, is there a place to document it and suggest best practices?

Koopa1018:
Discovered a weird bug with `Interactable`. If a child class is set to `tool`, then `sprite` will run its `onready` assignment code anytime a scene with the child class is loaded (when the child runs `_ready()`)--and because the child class's hierarchy doesn't exist in the editor, the `onready` code is liable to set the child's `sprite` field to Unassigned (and cause runtime problems) the moment you open the scene for editing.

This makes it a little impractical to use editor-facing `setget` functions in child classes of `Interactable`, so for now, I've just removed the `tool` flag from doors and made the `setget`s part of `_ready()` instead.

**To reproduce:**

- Add the `tool` keyword to the top of `door.gd`
- Add a door to any scene
- Close and reopen the scene in the editor
- Check the door's Sprite box in the inspector.

Koopa1018:
Tbh I was just making it a toolscript so I could load the DoorSkin animation data offline, rather than at runtime, which really doesn't seem necessary 😅 

Been wondering if that was the plan for the LD. Glad to finally know 🙂 

Hah! I've had that very problem with dialog windows myself~

GTcreyon:
What's the current state of this? Just wondering if there's anything that needs to be finished up.

Koopa1018:
One last check and no, it doesn't appear there's anything left. Ready to merge if you are.

GTcreyon:
A very good PR, that you should be proud of! ^^

Out of curiosity, do you plan on contributing anything else to the project in future?

Koopa1018:
I certainly do! Currently planning to upgrade pipes next, now that `InteractableWarp` is a reality. Paintings would also be a good target (they're basically doors with a different entry animation), except that they would also require new art assets (bare minimum, a picture frame and Mario jumping while facing back).

I feel like these aren't especially high-priority, but I have no clue what I can do for the level designer, so this is the best I've got 😅

GTcreyon:
Would you be interested in joining the development team directly? We might be able to keep you updated more accurately that way.

Re; paintings, there's already a few painting sprites in the game files. You can see them on the shine map in the pause menu.

Koopa1018:
I'd love to!

Man, I really gotta look through all the files sometime :)

GTcreyon:
Great! Do you have a Discord account? That's our primary communication method. My tag is creyon#1828.

If not, you're welcome to provide alternative contact methods.

Koopa1018:
I'm Sirius Bizdness#5554.

PR #29, opened 2022-11-25
Refactor pipes again

Koopa1018:
**Plans for PR:**
- [x] Now that the door refactor has landed, it's now possible to make pipes extend `InteractableWarp`. This makes the code much smaller and more intelligible, though it does also require rebuilding the prefab to put the Area2D as the root and the StaticBody2D as a child.
- [x] Like doors, move the pipe origin to the bottom to facilitate snapping to ground.
- [ ] Quick pipe--ground pound to enter pipes fast.
- [ ] Full visual support for lightly rotated pipes, as demonstrated by Wobuffet Forme Pikachu on the Discord. Current code always orients and slides the player as if all pipes are exactly vertical.
- [ ] Maybe support for pipes at _any_ angle, including sideways and upwards? I can think of exactly one place in the main campaign that this might _possibly_ be useful (SL igloo entrance), but it'd be lots of fun in user-created levels.

**Questions on current status:**

- Knowing the level designer is meant to be the tool of choice for the main campaign, is the origin change needed? Would it be better to revert that part?
- What the hey is going on in the disable code? It's not used right now, but I don't want to remove it until I know if the collision mask, `ride_area`, etc. are important.
- Similarly, is it necessary to manually switch player state/animation to neutral on warp end? Seems like it works fine to let the player object self-regulate.

**Notes on future plans:**
I would like to implement quick-pipe, but in the current codebase there's no elegant way to trigger an Interactable based solely on player state.

My plan was to override Pipe's `_interact_check()` function so it checks if down is pressed _or player is pounding_, but unfortunately there's no way to get player state in `_interact_check()` at the moment--`InteractableWarp` does store a reference to the player, but it's not assigned until _after_ `_interact_check()` runs in the base `Interactable` method. I really, really do not want to modify `Interactable` to make this one case work--not only is that outside the scope of a pipe PR, but it seems stupid to make `_interact_check()` functionally the same as `_state_check(body)` when they're clearly intended to do different things!

I can think of a hacky way to do it: have `_interact_check()` always return `true` and move input checking into `_state_check(body)` instead. It's a terrible solution--relies on these two functions being used solely as two sides of an `and` operation in `Interactable`, which means it could break if `Interactable`'s implementation changes.
**Would it be best to just do it the hacky way** (preferably with clear explanations as to what's going on and why it's a bad solution)**?**

### DIFF
--- a/classes/interactable/interactable_warp.gd
+++ b/classes/interactable/interactable_warp.gd
@@ -3,27 +3,38 @@ extends Interactable
 # A variant of Interactable that warps the player somewhere,
 # either to another scene or someplace in the current one.
 # 
-# To implement this base class, there are two functions to extend:
-# - _animation_length() returns the expected duration of the animation
-#   as an int.
-# - _update_animation(_frame, _player) updates the enter animation for the
-#   given frame.
-# For further customization, you can override _interact_check() if you want
-# it to respond to a different button than "up",
-# or override _begin_scene_change(target_pos, scene_pos) to change what
-# transition will be used on scene change (star iris out by default).
-# Additionally, override _exit_pos_offset() to shift all destination positions
-# across the game by some amount. The value returned will be added to the 
-# destination.
-# 
-# Please note that if the particular warp is set to move to a different scene,
-# the exit transition will begin TRANSITION_SPEED_IN frames before the end of
-# the animation.
-#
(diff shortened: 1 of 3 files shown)


Koopa1018:
Guess I don't need to use PR's anymore :/

Issue #24, opened 2022-11-24
Cherry clones do not spray water with FLUDD


ChaseyBOIDude:
**Describe the bug**
Mario's Clone Spray's Water while Mario Dosen't.

**To Reproduce**
Press "/" and type in "cherry", then get the FLUDD and press "c".

**Expected behavior**
Both Mario and the Clone are going to spray water.
[Screen recording 2022-11-24 10.08.48 AM.webm](https://user-images.githubusercontent.com/118941203/203817509-b61f4956-eafc-4d14-9226-c9c38191b100.webm)

**System Information**
 - OS/Browser: Chrome
 - Game version: 0.1.5
 - Other specs:

**Additional context**
None

GTcreyon:
> Would be closed by #16.

It was not, actually. This issue is more complex.

Koopa1018:
Working on `bubbles.gd` to try and solve the problem.

Question: why does `hud` call `refresh()` on the bubbles in its process function, rather than `bubbles` doing that itself?

GTcreyon:
> Working on `bubbles.gd` to try and solve the problem.
> 
> Question: why does `hud` call `refresh()` on the bubbles in its process function, rather than `bubbles` doing that itself?

I don't think there was a reason for it. Ask @jaschutte.

PR #39, opened 2023-01-16
Player decentralizing

### LINKED ISSUE #16: Properly decentralise player instances
Currently, a lot of player properties such as HP and FLUDD water are stored in the Singleton script. This is unhelpful because it makes it hard to create additional instances of the player for things like cherry clones and multiplayer. It's also generally bad practice to use a global script in this way.

Not immediately.

**Additional context**
If it were to be done, some rewrites to how the HUD works would be needed, to ensure they track info correctly.

Koopa1018:
Attempting to resolve #16. See the issue for details.

Most changes worth noting are in `player.gd`, `singleton.gd`, `inter_scene_data.gd`, and `death_manager.gd`. Everything else is just updating changed APIs.

Some implementation quirks worth noting:

1. Per-player data is now centralized to the player instance, but it still has to be copied into the singleton during scene changes. I created `InterSceneData` to make this easy to do, as well as easy to replicate when each player needs its own data to persist.
2. `Singleton.warp_to(...)` now stores inter-scene data, rather than relying on the calling code to remember that step. As a side effect, it now accepts a `PlayerCharacter` object (`player.gd`) to copy data from, though `null` can be passed if data should not be saved (or if no player exists, as on the title screen). 
3. Target-scene warp position is _not_ part of `InterSceneData`, since it's really a separate thing from how much water, health, etc. a player has. It can still be set using `Singleton.warp_to(...)`, but the current codebase doesn't do this, and sticks to the old way.
4. `death_cover.gd` has been reworked into `DeathManager`. Originally, death cover began its fade-and-reload sequence by checking the player's death state each frame, but doing this now requires a player reference--and we can't just get one using `onready var player = $"/root/Main/Player"` because the death cover is created on the title screen, which has no Player.
The solution I went with is to have the player object call an "I'm dead" function on the death cover, then just track whether that function's been called. This function accepts a `PlayerCharacter` as well--the plan is to eventually have the death cover track multiple players' death states. This track-multiple-players plan is part of why the script was renamed.
The other reason is the Single Responsibility Principle. Objectively, fading the screen should be a different system from death. Because of this, I've made the death cover a _child_ of `DeathManager`, rather than being one and the same. Ideally the cover would handle fading out on its own, but as that would suggest a full rewrite of the transition system, I'm leaving the actual fade in the hands of `DeathManager` for now.

On testing, it seems this does _not_ affect #24.


GTcreyon:
This isn't a major issue and doesn't necessarily need to be fixed in this PR, but take note that cherry clones seem to behave weirdly on semisolids. If you duck and then unduck, they will teleport upwards slightly. I assume this is due to a mis-use of the `dive_correct()`  function, but I'm not sure.

Aside from that, I think this all looks good? I like the way warps and such are handled.

classes/console/console.gd
```diff
@@ -41,8 +41,8 @@ func run_command(cmd: String):
 				if path == "":
 					Singleton.log_msg("Scene does not exist.", Singleton.LogType.ERROR)
 				else:
-					Singleton.set_location = null
-					var err = Singleton.warp_to(path)
+					Singleton.warp_location = null
+					var err = Singleton.warp_to(path, $"/root/Main/Player")
```
GTcreyon, line 45: In future, if we go ahead with multiplayer and such, then we'd need to replace these absolute paths with some kind of reference, but this works.

classes/global/singleton/inter_scene_data.gd
```diff
@@ -0,0 +1,23 @@
+class_name InterSceneData
+extends Reference # Reference counted because where to consistently free an Object?
```
GTcreyon, line 2: Reference seems correct.

classes/pickup/coin/coin_pickup.gd
```diff
@@ -21,10 +21,10 @@ func _ready_override() -> void:
 		vel.y = -7 * 0.53
 
 
-func _add_coins(num: int) -> void:
+func _add_coins(num: int, player) -> void:
```
GTcreyon, line 24: Consider adding a type hint to `player` here?

classes/player/player.gd
```diff
@@ -1,3 +1,4 @@
+class_name PlayerCharacter
```
GTcreyon, line 1: I think just `Player` would be fine? I'm surprised it wasn't already called that, actually. PlayerCharacter is nice too, though. Your call.

gui/timer/timer.gd
```diff
@@ -45,18 +45,12 @@ func format_time(overall_seconds):
...
+		Singleton.warp_location = Vector2(110, 153)
 		FlagServer.reset_flag_dict()
-		Singleton.warp_to(RESET_SCENE_PATH)
-		var player = get_node_or_null("root/Main/Player")
-		if player != null:
-			player.position = Vector2(110, 153)
+		Singleton.warp_to(RESET_SCENE_PATH, null)
 	
 	rect_scale = Vector2.ONE * max(floor(OS.window_size.x / Singleton.DEFAULT_SIZE.x), 1)
 	if !Singleton.meta_paused and running:
```
GTcreyon, line 56: We should probably rewrite this part entirely, but, that's for a different PR.

GTcreyon (approved):
All looks good!

classes/player/player.gd
jaschutte: I think PlayerCharacter is fine considering this is about the character, we can use the term Player for the user's account in the future (if we add such a thing). I generally prefer specific names over generic ones.

jaschutte (approved):
Everything looks good to me

classes/console/console.gd
Koopa1018: Oh yes, definitely.

classes/pickup/coin/coin_pickup.gd
Koopa1018: Ahhh, yep, good call

PR #44, opened 2023-01-19
Quieter FLUDD box

Koopa1018:
FLUDD box opening was painfully loud by default. Dropped the volume a bit.


GTcreyon (approved):
I'm not sure it's necessary to have a pull request for single-line or even single-commit changes like this. That said, this seems fine.

PR #45, opened 2023-01-19
Bob-omb sound effects

Koopa1018:
Implements fuse burning sound, hit-with-spin sound (definitely needs review), and a rising buildup to explosion. Also trims the explosion sound so it fits snugly next to the fuse and buildup.

@brickblock369 wasn't fond of the buildup, suggested a sine-wave beep instead. Thoughts on that?


classes/entity/enemy/bobomb/bobomb.gd
```diff
@@ -6,15 +6,30 @@ extends EntityEnemyWalk
 # When their fuse is lit, they do not lose focus on the player.
 # They can be struck to send them flying a short distance before exploding.
 
+# When two bombs play the buildup sound at once, the sound gets confusing.
+# Would be nice to have it choose from two or three different pitches so
+# the player can differentiate which sound belongs to which bomb, but
```
GTcreyon, line 11: Perhaps having a bob-omb flash red in time with the sounds might help? That wouldn't solve this directly, but would communicate the same information.

GTcreyon (approved):
I think this is all fine as-is, the knock sound is good in my opinion.

You may wish to add the red flash effect as mentioned in a comment, but you can do that in a separate PR if preferred.

classes/entity/enemy/bobomb/bobomb.gd
Koopa1018: Good thought; definitely for a separate PR though 😛

PR #47, opened 2023-01-19
Heftier ground pound feedback

Koopa1018:
**NOTE: this branch is based on #46. Probably best to review that one first, because this one can't automerge with that one yet.**

Refines the in-air spin before the pound begins, and adds screen shake on impact. (May be best if screen shake could be disabled in Options, but I'm not sure how to implement that.)

Ended up adding a little extra time from button press to pound beginning. The little extra hang time helps the spin feel slower, which means the drop doesn't feel slow by comparison. Tried my hardest to get it to look good without adjusting actual hang time, but them's the breaks I guess :/

I ended up adding an extra pose to Mario's sprite sheet. It's unused, but it's there still. Should I remove it?


classes/dejitter_group/dejitter_group.gd
```diff
@@ -18,6 +20,7 @@ func _physics_process(_delta):
 			offset_vec = Vector2.ZERO
 		else:
 			offset_vec = carrier_node.texture.get_size().posmod(2) / 2
+	offset_vec += dejitter_position
```
GTcreyon, line 23: A comment here might help. Not sure what this does.

project.godot
```diff
@@ -665,6 +665,10 @@ locale_filter=[ 0, [  ] ]
 
 project/assembly_name="Super Mario 63 Redux"
 
+[network]
+
+limits/debugger_stdout/max_chars_per_second=2147483647
```
GTcreyon, line 670: Was this accidental? I don't see what relevance this bears. I also don't know what this does. :P

Koopa1018: That...should already be in master. It raises the debug output limit to its absolute maximum, so it's possible to print_debug(...)` every frame and the console doesn't lock up within a few seconds.

classes/dejitter_group/dejitter_group.gd
Koopa1018: I can't remember either. Time for some code archaeology ⛏️

Koopa1018: Done.

...actually, I went ape and commented the entirety of `DejitterGroup`, including a brief docstring.

Koopa1018:
- Extra frame removed from Mario's sprite sheet.
- Reverted out-of-place `project.godot` change.
- Commented pretty much all of `dejitter_group.gd`, including the spot you requested.
- Resolved merge conflicts to the best of my ability.

PR #54, opened 2023-01-22
Customizable lifetime for falling logs

GTcreyon:
# Description of changes
Logs now have a customizable lifetime, set with an export variable. They will fall until they reach a max speed (such that Mario doesn't fall off), and fade out when the timer runs out.


Koopa1018 (approved):
Looks good to me. You want to wait for Charlotte's review, or not so much?

PR #57, opened 2023-01-24
Warp command refactor

Tryzen-0:
# Description of changes
- Optimized warp command for tutorial_1, Now the command goes as follows:
`w tutorial_1 <number from 1 to 4 here>`
- I also added castle lobby warp
`w castle_lobby`

- I made this to make myself feel useful and to possibly help out and learn a bit of coding.

EDIT: this is not more optimized just better maintainable, my bad!

### DIFF
--- a/classes/console/console.gd
+++ b/classes/console/console.gd
@@ -28,14 +28,13 @@ func run_command(cmd: String):
 					Singleton.log_msg("No second argument.", Singleton.LogType.ERROR)
 					return
 				match args[1]:
-					"t1":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_1.tscn"
-					"t1r2":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_2.tscn"
-					"t1r3":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_3.tscn"
-					"t1r4":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_4.tscn"
+					"tutorial_1":
+						if len(args) == 2:
+							Singleton.log_msg("No scene number provided for this course.", Singleton.LogType.ERROR)
+							return
+						path = "res://scenes/levels/tutorial_1/tutorial_1_" + args[2] + ".tscn"
+					"castle_lobby":
+						path = "res://scenes/levels/castle/lobby/castle_lobby.tscn"
 					_:
 						path = ""
 				if path == "":
--- a/classes/console/console.gd
+++ b/classes/console/console.gd
@@ -29,7 +29,7 @@ func run_command(cmd: String):
 					return
 				match args[1]:
 					"t1", "tutorial_1":
-						if len(args) == 2:
+						if len(args) <= 2:
 							path = "res://scenes/levels/tutorial_1/tutorial_1_1.tscn"
 						else:
 							path = "res://scenes/levels/tutorial_1/tutorial_1_" + args[2] + ".tscn"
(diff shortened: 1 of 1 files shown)


classes/console/console.gd
```diff
@@ -28,14 +28,13 @@ func run_command(cmd: String):
...
-					"t1r3":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_3.tscn"
-					"t1r4":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_4.tscn"
+					"tutorial_1":
+						if len(args) == 2:
+							Singleton.log_msg("No scene number provided for this course.", Singleton.LogType.ERROR)
+							return
+						path = "res://scenes/levels/tutorial_1/tutorial_1_" + args[2] + ".tscn"
+					"castle_lobby":
```
GTcreyon, line 36: Have an option for just `"lobby"`.

classes/console/console.gd
```diff
@@ -28,14 +28,13 @@ func run_command(cmd: String):
...
 				match args[1]:
-					"t1":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_1.tscn"
-					"t1r2":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_2.tscn"
-					"t1r3":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_3.tscn"
-					"t1r4":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_4.tscn"
+					"tutorial_1":
```
GTcreyon, line 31: Perhaps have the option of using the shorter name `"t1"`?

You can do this with the line `"t1", "tutorial_1":`. This will allow both as options.

classes/console/console.gd
```diff
@@ -28,14 +28,13 @@ func run_command(cmd: String):
...
+						if len(args) == 2:
```
GTcreyon, line 32: Consider making it default to room 1 when no second argument is given?

GTcreyon (changes requested):
A nice simple pull request. A good start, but a few things that I would change before it can be merged.

classes/console/console.gd
Tryzen-0: should be fixed now! check the 2nd commit.

GTcreyon: Nice!

classes/console/console.gd
```diff
@@ -28,14 +28,13 @@ func run_command(cmd: String):
...
-					"t1":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_1.tscn"
-					"t1r2":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_2.tscn"
-					"t1r3":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_3.tscn"
-					"t1r4":
-						path = "res://scenes/levels/tutorial_1/tutorial_1_4.tscn"
+					"t1", "tutorial_1":
+						if len(args) == 2:
```
GTcreyon, line 32: I would change this to `if len(args) <= 2:` to be clearer on what it's checking, even if the functionality is identical.

classes/console/console.gd
```diff
@@ -28,14 +28,13 @@ func run_command(cmd: String):
...
+							path = "res://scenes/levels/tutorial_1/tutorial_1_1.tscn"
+						else:
+							path = "res://scenes/levels/tutorial_1/tutorial_1_" + args[2] + ".tscn"
```
GTcreyon, line 35: Maybe check if that scene at that path actually exists, so the player can't crash the game with `w t1 somethingrandom`.

GTcreyon (changes requested):
Getting there, these are some good improvements.

classes/console/console.gd
Tryzen-0: For some reason it doesn't crash the game and I don't know why.
I will have to figure out how to check if it exists cuz I don't know how that works yet.

Tryzen-0: ok fixed

GTcreyon: I realize now that there was actually already a fallback in place for it, but yours is better, I think.

GTcreyon (approved):
All good!

PR #58, opened 2023-01-26
Optimized if-file-exists check on warp command

Tryzen-0:
# Description of changes
Removed 1 if loop.

### DIFF
--- a/classes/console/console.gd
+++ b/classes/console/console.gd
@@ -37,19 +37,16 @@ func run_command(cmd: String):
 						path = "res://scenes/levels/castle/lobby/castle_lobby.tscn"
 					_:
 						path = ""
-				if path == "":
+				var err = OK
+				if path == "" or not File.new().file_exists(path):
 					Singleton.log_msg("Scene does not exist.", Singleton.LogType.ERROR)
 				else:
 					Singleton.warp_location = null
-					var file_check = File.new()
-					if file_check.file_exists(path):
-						var err = Singleton.warp_to(path, $"/root/Main/Player")
-						if err == OK or err == null:
-							Singleton.log_msg("Warped to " + path)
-						else:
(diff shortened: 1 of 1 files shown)


classes/console/console.gd
```diff
@@ -37,19 +37,16 @@ func run_command(cmd: String):
 						path = "res://scenes/levels/castle/lobby/castle_lobby.tscn"
 					_:
 						path = ""
-				if path == "":
+				var err = OK
+				if path == "" or not File.new().file_exists(path):
```
jaschutte, line 41: Is this the correct function?
Shouldn't ResourceLoader.exists() be used instead? Considering scenes are resources and godot might remove the source in the export.

GTcreyon: Yeah, `ResourceLoader.exists()` is better. Using the `File` class on internal resources won't work when running the exported executable. Where did the source use `File.new().exists()`?

jaschutte: > Yeah, `ResourceLoader.exists()` is better. Using the `File` class on internal resources won't work when running the exported executable. Where did the source use `File.new().exists()`?

Lines 44 & 45

GTcreyon: I think that was something that Tryzen added but I forgot to check last time.
However, I don't think it was her fault. It turns out that the `scene` command has been using it, but nobody noticed because nobody uses that command in the exported version (it doesn't make much sense without context).

...Actually, that command doesn't work *at all*, because it requires the use of `/` for the scene paths. And you can't type that into the console, because it just closes when you try to. Whoops.

GTcreyon, line 41:
```suggestion
				if path == "" or not ResourceLoader.exists(path):
```

GTcreyon (changes requested):
Apologies for taking so long to get to this! A slight change needs to be made, but otherwise good ^^

GTcreyon (approved):

PR #66, opened 2023-02-11
Enhance triple jump anim

Koopa1018:
# Description of changes
Updates Triple Jump animation to look a bit more centered, and adds a moment at the start where Mario is not in the flip pose so it looks a tad less abrupt.

Also a slight refactor to backflip animation code, because I thought I'd be updating that animation as well, but. Turns out backflips look fine as they are 😕 so I guess it's just a little refactor for consistency.

# Reviewer(s)
@GTcreyon @Charpurrr 

# Issue(s)
N/A

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -805,6 +805,7 @@ func manage_triple_flip() -> void:
 			dir = -1
 		var multiplier = 1
 		if current_nozzle == Singleton.n.none:
+			# Flip slower if wearing FLUDD
 			multiplier = 2
 		sprite.rotation = dir * multiplier * TAU * ease_out_quart(float(triple_flip_frames) / TRIPLE_FLIP_TIME)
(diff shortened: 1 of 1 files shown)


classes/player/player.gd
```diff
@@ -287,8 +287,10 @@ func player_physics():
 	manage_invuln()
 	manage_buffers()
 	manage_dive_recover()
-	manage_triple_flip()
-	manage_backflip_flip()
+	if state == S.TRIPLE_JUMP:
```
GTcreyon, line 290: In future it might be worthwhile taking code for stuff like this and moving it into child nodes to form a more state-machine-like approach, but this is fine for now.

GTcreyon (approved):

classes/player/player.gd
Koopa1018: Oh yeah, that'd definitely be better. This whole script is a huge mess 😆

PR #67, opened 2023-02-11
Enhance spin animation

Koopa1018:
# Description of changes
Adds two new spin animation phases: one frame of extra-fast spinning at the start, and a slow eight-frame loop after the existing animation has gone for a moment. Slow animation begins quite fast, but quickly slows down to a speed similar to the Flash version. Additionally, slow animation has its arms fully extended for maximum helicopter resemblance.

Of note: the new animation has frames 32px wide. All the rest of Mario's poses are 24px wide.

Work known needed:
- [x] ~~FLUDD doesn't integrate with the new animations. Need to code that, and possibly draw new frames.~~ Added, along with a new sync-FLUDD-sprite-with-player system that should be much more flexible than just hard-coding everything.
- [x] ~~Charlotte requested that the original and fast-spin animations be updated to match slow-spin's T-pose. Whether or not the T-pose is chosen, the animations definitely should be consistent.~~ Shima is redrawing the Mario sprites, so this is kind of a moot point.

# Issue(s)
N/A


GTcreyon:
Consider merging from `main` and resolving conflicts, due to changes in #82.

Koopa1018:
Will do

Koopa1018:
Considering Shima is redrawing player sprites from scratch, I might as well give up on updating the spin sprites and mark this one ready for review

GTcreyon:
> Considering Shima is redrawing player sprites from scratch, I might as well give up on updating the spin sprites and mark this one ready for review

Is merging this still useful? Will it cause conflicts?

Koopa1018:
At this point there probably will be conflicts, but this does add in a couple useful things:

1. A prebuilt multi-phase spin animation, including speed modulation for smooth animation transitions
2. Perhaps more importantly, a framework for easy setting of FLUDD angles for specific player animation frames, e.g. "On frame X of Mario's animation Y, FLUDD should be in position Z."

GTcreyon:
> At this point there probably will be conflicts, but this does add in a couple useful things:
> 
>     1. A prebuilt multi-phase spin animation, including speed modulation for smooth animation transitions
> 
>     2. Perhaps more importantly, a framework for easy setting of FLUDD angles for specific player animation frames, e.g. "On frame X of Mario's animation Y, FLUDD should be in position Z." Should be way easier to use than hard-coding behavior for every animation involved.

These are valid points. The FLUDD angle part is particularly valuable. However, the changes to the spin animation will inevitably be overwritten, because my branch moves animation handling into a different script entirely. Would it perhaps be worthwhile closing this PR, but moving some of your additions to the other branch?

Koopa1018:
Seems like a good idea.

GTcreyon:
> Seems like a good idea.

The branch in question is enhance-mario-sprites. I don't think I'll be working on it any more today, so you're welcome to fiddle with it if you'd like.

PR #69, opened 2023-02-21
Make "underwater effect" bypassable

Koopa1018:
# Description of changes
For ease of future sound work, refactors the global audio effect when player is underwater.

The "underwater effect" is made up of two effects, a low-pass filter and a reverb. Before this PR, each of these were applied to the master bus directly. This PR moves them each to their own dedicated buses, taking care to preserve effect order. The overall effect is identical, and no sound effects have been changed.

The refactor allows for some sounds to skip one or both of the effects, which should come in handy for future UI sounds (no effects) and sounds specifically designed to play underwater (reverb only).

# Reviewer(s)
@GTcreyon 
@brickblock369 
@Olimus192
lmk if there's anyone else to ping
# Issue(s)
N/A


default_bus_layout.tres
```diff
@@ -1,41 +1,59 @@
...
+bus/4/send = "~~Water Muffle:"
+bus/5/name = "Coin"
+bus/5/solo = false
+bus/5/mute = false
+bus/5/bypass_fx = false
+bus/5/volume_db = 0.0
+bus/5/send = "SFX"
+bus/5/effect/0/effect = SubResource( 3 )
+bus/5/effect/0/enabled = true
+bus/6/name = "~~SFX Clear In Water~~"
```
GTcreyon, line 54: I'm not sure why these tildes are here. It's not a problem, but, what exactly do they communicate?

GTcreyon (approved):
All good, but see my comment on what the tildes are for.

default_bus_layout.tres
Koopa1018: Hmmm. That's a fair point.
I think I was trying to communicate the hierarchy of the bus layout, but on reflection, I don't think that's necessary. Mostly future SFX placers just need to know "don't put stuff on the Water Muffle and Water Verb buses, those ones are special."

Issue #61, opened 2023-01-31
Question: What does /c do?


PoketrainerMS:
I was looking through the code for the console commands, and I noticed `/c` is a command, and it toggles some kind of "classic mode", though when I tried it, nothing seemed to change.
What does/will it do?

###### also how do i add labels to an issue?

ChaseyBOI801:
I think it changes the sprites from the new ones to the ones from the original

PoketrainerMS:
> I think it changes the sprites from the new ones to the ones from the original

That's what i thought, but no.

GTcreyon:
It does nothing at all.

It was originally planned to change some things about the physics to make it more accurate to the original, but that idea was scrapped.

PoketrainerMS:
Interesting.

ChaseyBOI801:
Maybe we could bring back /c? sounds like it would have been cool.

GTcreyon:
> I know Wario is Going to be in the Game so maybe for the `/c` it changes his sprite to the Wario Sprite from Wario Land 4? idk just an idea.

We will not be adding sprite swaps of any kind. Too much effort for something that should be relegated to a texture mod.

ChaseyBOI801:
> > I know Wario is Going to be in the Game so maybe for the `/c` it changes his sprite to the Wario Sprite from Wario Land 4? idk just an idea.
> 
> We will not be adding sprite swaps of any kind. Too much effort for something that should be relegated to a texture mod.

Maybe We could add it after the game releases and make it part of a update.

GTcreyon:
> Maybe We could add it after the game releases and make it part of a update.

No. We would be focusing on other things. Again, it would have to be a mod made by someone else.

Locking this thread now, since it has become off-topic.

PR #71, opened 2023-02-22
Implement /c

ChaseyBOI801:
I did the "/c" code because you said no, I don't have the classic Mario sprite sheet yet but it's coming soon.

# Description of changes

/c command changes mario's sprites to the original SM63 sprites
# Reviewer(s)

 @GTcreyon

### DIFF
--- a/classes/console/console.gd
+++ b/classes/console/console.gd
@@ -74,7 +74,33 @@ func run_command(cmd: String):
 				Singleton.log_msg("Water refilled.")
 			"c":
 				Singleton.classic = !Singleton.classic
-				if Singleton.classic:
+				extends Node
+
+                                var sprite_folder = "classes"
+
+                                func _ready():
+                                print("Ready to import sprites")
+
+                                func _process(delta):
+                                if Input.is_action_pressed("ui_accept"):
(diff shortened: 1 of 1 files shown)


GTcreyon:
Look. I've said before in issue #61 that sprite swaps will not be added, and `/c` is not used anymore. If you want to argue with that decision, you should open a separate issue and explain why you think that, but I doubt it will change much. We don't want to add any of these features, nor to include any code by the community that adds them. This should only ever be added in a mod, not in the base game. As such, this pull request will not be merged.

That aside, the code you have given us will not run. The indentation isn't right, and it includes function declarations like `func _ready():` in the middle of code blocks, as well as `extends Node`, which should only ever be at the top of the script. It seems like this happened after you pasted it in from Discord, but the code you posted there - while formatted better - also would not work. I'm not sure if it will run, but even if it does, it won't do what you want it to do. It's important to test your code before opening a Pull Request.

While I appreciate that you want to help out the game's development, doing and expecting things that you've been told multiple times will not and cannot happen is wasting your time, and our time too. It's also sad to have to stomp on your expectations - I don't like having to do that.

I'll be closing this PR now. It will not be added, even if the code itself is fixed.

PR #74, opened 2023-02-28
Fix fludd meter not being visible

GTcreyon:
Fixes an issue where the FLUDD meter is not visible under any circumstances.


jaschutte (approved):
Such a code breaking change, how could we possibly merge this into main?!

PR #88, opened 2023-03-03
Align hover spray to nozzle

### LINKED ISSUE #81: Misaligned hover spray
**Describe the bug**
The water that comes out of the hover FLUDD is not properly aligned to the nozzle. It is offset slightly up-left.

**To Reproduce**
Use the hover fludd under any conditions.

**Expected behavior**
Water spray should neatly fit into the nozzle part of the sprite.

**Screenshots**
image

**System Information**
 - OS/Browser: Linux
 - Game version: v0.2.0.alpha

Koopa1018:
# Description of changes
Adjusts hover spray offset to visually match the position of the nozzle.

This was done by a simple change to the particle system's hard-coded position offset, as well as some refactoring to make the offset easier to tune. The spray plume effect was also being placed using this offset, and it looked fine already, so I had to give it its own offset value to keep it where it already was.

While testing, I discovered that not only did the spray effect have a different offset when diving, but that said offset didn't even change with the player's rotation. By making the offset rotate appropriately, I was able to completely remove the dive offset and rely on the normal offset value for everything, and end up with spray consistently coming from the nozzle no matter what angle Mario is currently diving.

# Reviewer(s)
@GTcreyon @RandomCatDude @jaschutte 

# Issue(s)
Closes #81

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -420,6 +420,8 @@ func adjust_swim_x() -> void:
 	vel.x += (swim_adjust - vel.x) * FPS_MOD
 
 
+const SPRAY_ORIGIN_STAND = Vector2(-10, -2)
+const SPRAY_ORIGIN_DIVE = Vector2(1, -9)
 var hover_sound_position = 0
 var nozzle_fx_scale = 0
 func fixed_visuals() -> void:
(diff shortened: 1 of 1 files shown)


GTcreyon:
Make sure to actually request reviews instead of just a mention.

classes/player/player.gd
```diff
@@ -420,6 +420,8 @@ func adjust_swim_x() -> void:
 	vel.x += (swim_adjust - vel.x) * FPS_MOD
 
 
+const SPRAY_ORIGIN = Vector2(-9, 4)
```
GTcreyon, line 423:
```suggestion
const SPRAY_ORIGIN = Vector2(-9, 6)
```
This allows it to fit snugly into the nozzle without being visible above it.

GTcreyon (changes requested):
This is really good, I like the refactor. Just a small change to be made regarding the offset chosen.

classes/player/player.gd
Koopa1018: Ah yep, that does look better!

Koopa1018:
> Make sure to actually request reviews instead of just a mention.

Wait those are separate? Dang, I had no idea. I'll keep that in mind.

(Wait but. If that doesn't ask for review, what's the Reviewers section of the template for?)

GTcreyon (approved):

GTcreyon:
> (Wait but. If that doesn't ask for review, what's the Reviewers section of the template for?)

Nothing. I think @jaschutte added it thinking it would, but it doesn't. I'll probably remove it.

Issue #35, opened 2023-01-09
Missing Sounds Masterlist


Koopa1018:
In the current release, the following gameplay events don't have sound, even though they should.
##### Mario
- [ ] Walk on wood*
- [ ] [SUGGESTED] Jump, in addition to the existing voice clips
- [ ] Ground pound flip
- [ ] Ground pound impact
- [ ] Backflip
- [ ] Spin in air
- [ ] [OPTIONAL] Spin in water
- [ ] Swim
- [ ] Rocket nozzle charge
- [ ] Rocket nozzle launch

*has sound, but could use a more purpose-built one

##### Enemies
- [ ] Goomba death
- [ ] Goomba knockback
- [ ] Bob-omb fuse
- [ ] Bob-omb buildup (drafted but not approved, second draft pending)
- [ ] Bob-omb knockback
- [ ] Cheep-Cheep knock/death (one action)
- [ ] [OPTIONAL] Goonie flap
- [ ] [OPTIONAL] Goonie stepped on

##### Items
- [ ] Water bottle collect, small
- [ ] Water bottle collect, big
- [ ] Red coin
- [ ] [OPTIONAL] Blue coin

##### GUI
- [ ] Pause game
- [ ] Button mouse over
- [ ] Button click
- [ ] [OPTIONAL] Dialogue box type
- [ ] Recover health

**Do you intend to contribute this yourself? If so, how do you plan on doing it?**
As the team sound designer, I'll definitely be chipping away at this over time.

PR #89, opened 2023-03-04
Add sound infrastructure to pickup baseclass

Koopa1018:
# Description of changes
Adds sound support directly to the pickup base class, so it can be shared between all pickups with zero effort.

The design is similar to #85, just a whole lot simpler (only one sound hook instead of three).

# Reviewer(s)
@GTcreyon @jaschutte

# Issue(s)
Important step towards #35.

### DIFF
--- a/classes/misc/residual_sfx/residual_sfx.gd
+++ b/classes/misc/residual_sfx/residual_sfx.gd
@@ -28,9 +28,11 @@ static func new_from_existing (
 	scene_root: Node,
 	destroy_on_finished = true
 ):
-	# Reparent to scene root.
+	# Reparent to scene root, while preserving global position.
+	var sound_pos = sfx.global_position
 	sfx.get_parent().remove_child(sfx)
 	scene_root.add_child(sfx)
+	sfx.global_position = sound_pos
 	
 	# Make the sfx player destroy on playback (if desired)
 	if destroy_on_finished:
--- a/classes/misc/residual_sfx/residual_sfx.gd
+++ b/classes/misc/residual_sfx/residual_sfx.gd
@@ -19,3 +19,21 @@ func _on_ResidualSFX_finished():
 func _init(sound: AudioStream, pos: Vector2):
 	position = pos
 	stream = sound
+
+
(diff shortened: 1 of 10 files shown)


classes/misc/residual_sfx/residual_sfx.gd
```diff
@@ -19,3 +19,30 @@ func _on_ResidualSFX_finished():
...
+		_move_node_to(sfx, scene_root)
+		sfx.global_position = sound_pos
+	else:
+		# Reparent to the scene root.
+		_move_node_to(sfx, scene_root)
+	
+	# Make the sfx player destroy on playback (if desired)
+	if destroy_on_finished:
+		sfx.connect("finished", sfx, "queue_free")
+	# Lesgo
```
GTcreyon, line 43: I'll allow it.

classes/pickup/pickup.gd
```diff
@@ -65,6 +68,16 @@ func _pickup_id_setup() -> void:
 			queue_free()
 
 
+func _pickup_sound():
+	if sfx != null:
+		# Find an object we know will survive this object's destruction.
+		var safe_sfx_root = get_parent()
```
GTcreyon, line 74: I'm not so sure this is definitely "safe". I trust you, but I'm wondering why you chose to do this instead of parenting it to the scene root.

GTcreyon (approved):
I'm unsure of some of the decisions made here, but overall it's a good PR I think. Could do with some explanations though.

classes/pickup/pickup.gd
Koopa1018: Hmm. You know what, that's actually a better idea.

GTcreyon: I can revert the merge if needed.

GTcreyon: Nevermind, I can't easily do that. Best to just make a new PR.

Koopa1018: Might be a good idea

Koopa1018: ah yep, dang :/

PR #91, opened 2023-03-04
Water bottle SFX

Koopa1018:
**NOTE: Depends on #89.**

# Description of changes
Implements water bottle sound effects. One for small bottles, one for big.

Not sure if these should be spatialized or not. Currently they are, but I'm thinking maybe they shouldn't be?

# Reviewer(s)
@GTcreyon @brickblock369

# Issue(s)
Checks two boxes on #35


GTcreyon:
If this depends on #89, it might have made more sense to make it a branch off from that, rather than from `master`. But it's okay. I'll do that one first, then get to this one.

Koopa1018:
IIRC I did do that, and it just. didn't make a difference :/

GTcreyon:
> IIRC I did do that, and it just. didn't make a difference :/

Strange. Oh well.

GTcreyon (approved):

PR #114, opened 2023-03-18
Fixed pickup crash bug

### LINKED ISSUE #113: Respawned pickups crash game on collection
**Describe the bug**
Sometimes when the new_from_existing() function is called, an argument will be passed for the sfx parameter that's already been freed from memory. This crashes the game.

**To Reproduce**
Pick up water and wait for it to reload, then pick it up again.

**Expected behavior**
The pickup sound should be played. The game should not crash.

**Screenshots**
image

**System Information**
 - OS/Browser: Linux Ubuntu
 - Game version: Most recent
 - Other specs: Not applicable

**Additional context**
I'll try and work on this, it seems like the issue could be solved by checking if SFX is null when the argument is passed, rather than onready.

AidenBeresford:
Changed the Pickup class to contain a has_node() check in the _pickup_sound() method, and a check to ensure the object would not respawn before reparenting the AudioPlayer.

Closes #113

### DIFF
--- a/classes/pickup/pickup.gd
+++ b/classes/pickup/pickup.gd
@@ -96,7 +96,7 @@ func _pickup_sound():
 			ResidualSFX.new_from_existing(sfx, safe_sfx_root)
 		# Check to see if object should have an sfx, and if it does have an sfx.
 		elif sfx != null and !has_node(_sfx_path):
-			printerr("No SFXCollect Found :(")
+			printerr("This pickup should have SFXCollect, but it wasn't found. :(")
 	else:
 		sfx.play()
 
--- a/classes/pickup/pickup.gd
+++ b/classes/pickup/pickup.gd
@@ -54,7 +54,12 @@ func assign_pickup_id(id) -> void:
 
 func pickup(body) -> void:
 	_award_pickup(body)
-	_pickup_sound()
+	
+	if respawn_seconds == 0.0:
+		_pickup_sound()
+	else:
+		sfx.play()
+	
 	_pickup_effect()
 	_kill_pickup()
 	if persistent_collect:
@@ -85,7 +90,7 @@ func _pickup_id_setup() -> void:
 
 
 func _pickup_sound():
-	if sfx != null:
+	if sfx != null and has_node("SFXCollect"):
 		# Find an object we know will survive this object's destruction.
 		var safe_sfx_root = $"/root/Main"
 		# Anchor the sound source to that, then play it.
--- a/classes/pickup/pickup.gd
+++ b/classes/pickup/pickup.gd
@@ -54,12 +54,7 @@ func assign_pickup_id(id) -> void:
 
 func pickup(body) -> void:
 	_award_pickup(body)
-	
-	if respawn_seconds == 0.0:
-		_pickup_sound()
-	else:
-		sfx.play()
-	
+	_pickup_sound()
 	_pickup_effect()
 	_kill_pickup()
 	if persistent_collect:
@@ -90,11 +85,19 @@ func _pickup_id_setup() -> void:
 
 
 func _pickup_sound():
-	if sfx != null and has_node("SFXCollect"):
-		# Find an object we know will survive this object's destruction.
-		var safe_sfx_root = $"/root/Main"
-		# Anchor the sound source to that, then play it.
-		ResidualSFX.new_from_existing(sfx, safe_sfx_root)
+	# Check to make sure the object is killable.
+	if respawn_seconds == 0.0:
+		# If the has_node() check did not exist, it could pass an arg previously freed.
+		if sfx != null and has_node("SFXCollect"):
+			# Find an object we know will survive this object's destruction.
+			var safe_sfx_root = $"/root/Main"
+			# Anchor the sound source to that, then play it.
+			ResidualSFX.new_from_existing(sfx, safe_sfx_root)
+		# Some pickups i.e. coins, have no sfx at all. We need to run both checks.
+		elif sfx != null and !has_node("SFXCollect"):
+			printerr("No SFXCollect Found :(")
+	else:
+		sfx.play()
 
 
 func _kill_pickup() -> void:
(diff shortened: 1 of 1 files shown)


classes/pickup/pickup.gd
```diff
@@ -85,7 +90,7 @@ func _pickup_id_setup() -> void:
 
 
 func _pickup_sound():
-	if sfx != null:
+	if sfx != null and has_node("SFXCollect"):
```
Koopa1018, line 93: Not a good idea.
Best case scenario, `_sfx_path` is left default, `sfx` already references `SFXCollect`, this check is redundant.
Worst case scenario, `_sfx_path` has been changed, no `SFXCollect` node even exists in the object, check can never succeed, entire object is permanently silent.

classes/pickup/pickup.gd
```diff
@@ -54,7 +54,12 @@ func assign_pickup_id(id) -> void:
 
 func pickup(body) -> void:
 	_award_pickup(body)
-	_pickup_sound()
+	
+	if respawn_seconds == 0.0:
+		_pickup_sound()
+	else:
+		sfx.play()
```
Koopa1018, line 61: This `if` check should be inside `_pickup_sound()`, not outside it.

(`_pickup_sound()` is supposed to be the _entire_ sound-playing routine. Having sound-playing logic _outside_ it makes the codebase fragile.)

Koopa1018 (changes requested):

classes/pickup/pickup.gd
AidenBeresford: The worst case scenario of a permanently silent object is preferable. Without the has_node check, sfx can and has been passed to new_from_existing, causing a crash (since it's already been freed from memory). 

The has_node check acts as a failsafe to prevent sfx being passed, that way the worst possible outcome is a silent pickup.

classes/pickup/pickup.gd
AidenBeresford: Good point, I'll fix this in a second.

classes/pickup/pickup.gd
Koopa1018: That's fair.

In that case, try using `has_node(_sfx_path)` instead. That fixes both problems.

(And don't forget to comment your work, so future readers know what's going on here!)

Koopa1018: Wait--on second thought. 
Respawning is what causes the crash, right?
And with the other code change in place, this entire branch is unreachable when respawning is disabled, right?

So why do we even need to put a check here?

AidenBeresford: It's definitely pointless now, but that `respawn_seconds` check is kind of fragile. If someone changed the code in a way that allowed `sfx` to be reparented and `respawn_seconds` to not be 0.0, it would start crashing again. 

Anyways, I added a `printerr()` to the code, that way any future crashes related to sfx will alert us to it's cause immediately.

classes/pickup/pickup.gd
```diff
@@ -90,11 +85,19 @@ func _pickup_id_setup() -> void:
...
+	if respawn_seconds == 0.0:
+		# If the has_node() check did not exist, it could pass an arg previously freed.
+		if sfx != null and has_node("SFXCollect"):
+			# Find an object we know will survive this object's destruction.
+			var safe_sfx_root = $"/root/Main"
+			# Anchor the sound source to that, then play it.
+			ResidualSFX.new_from_existing(sfx, safe_sfx_root)
+		# Some pickups i.e. coins, have no sfx at all. We need to run both checks.
+		elif sfx != null and !has_node("SFXCollect"):
+			printerr("No SFXCollect Found :(")
```
Koopa1018, line 98: Don't...don't do this.

Save error messages for cases that _should not happen._ Items having no sfx node is intended behavior, and printing messages when intended behavior happens isn't usually helpful.

classes/pickup/pickup.gd
```diff
@@ -90,11 +85,19 @@ func _pickup_id_setup() -> void:
...
 func _pickup_sound():
-	if sfx != null and has_node("SFXCollect"):
-		# Find an object we know will survive this object's destruction.
-		var safe_sfx_root = $"/root/Main"
-		# Anchor the sound source to that, then play it.
-		ResidualSFX.new_from_existing(sfx, safe_sfx_root)
+	# Check to make sure the object is killable.
+	if respawn_seconds == 0.0:
+		# If the has_node() check did not exist, it could pass an arg previously freed.
+		if sfx != null and has_node("SFXCollect"):
```
Koopa1018, line 91: I still maintain that you shouldn't check for errors that are impossible.

Instead, leave a comment _pointing out_ that the error is impossible, and preferably _explaining why._

That way, future programmers aren't worried for no reason, and they know enough to leave it safe when they make changes.

Koopa1018 (changes requested):
This is a good step forward.

classes/pickup/pickup.gd
AidenBeresford: This is indeed a case that should never happen. If this happens, that check means `sfx` should exist, but can't be found.

AidenBeresford: I will improve my comments to be more clear.

classes/pickup/pickup.gd
AidenBeresford: Impossible error implies that it cannot happen under any circumstances. This error is not impossible, it's just not able to happen currently. The check is an insurance policy. I will add a comment better explaining it though.

classes/pickup/pickup.gd
GTcreyon: If I am understanding correctly, this error is printed when there is a sound to play, but no SFXCollect present to play it. In that situation, an error should be printed, no? So I don't see what @Koopa1018 is disputing.

classes/pickup/pickup.gd
GTcreyon: Indeed - including error checks to point out when code has been broken is useful, even if the code was originally bulletproof.

classes/pickup/pickup.gd
AidenBeresford: You are understanding correctly. I think the issue was my original comment, which was unclear on what the check actually does. This may have confused @Koopa1018 on the intent of the check.

Koopa1018 (approved):

PR #116, opened 2023-03-19
Fix bind reset not saving

GTcreyon:
# Description of changes
Fixes an issue where resetting keybinds to default does not result in changes being saved.

### DIFF
--- a/gui/pause/options/reset_binds.gd
+++ b/gui/pause/options/reset_binds.gd
@@ -2,5 +2,7 @@ extends TextureButton
 
 
 func _on_ResetBinds_pressed():
-	Singleton.load_input_map(Singleton.default_input_map)
+	var map = Singleton.default_input_map
+	Singleton.load_input_map(map)
+	Singleton.save_input_map(map)
 	Singleton.get_node("SFX/Back").play()


Koopa1018 (changes requested):
So glad this issue has been fixed! This caused problems for me when I was first starting out here.

On starting the game (at the title screen, like normal), the console gets flooded with this error message:
```
E 0:00:00.689   is_action_just_pressed: The InputMap action "timer_show" doesn't exist.
  <C++ Error>   Condition "!InputMap::get_singleton()->has_action(p_action)" is true. Returned: false
  <C++ Source>  main/input_default.cpp:104 @ is_action_just_pressed()
  <Stack Trace> timer.gd:44 @ _physics_process()
```

Apparently related, I am also unable to map the "show timer" action in the options menu--pressing a new button silently fails and leaves the input unmapped.

Switching Git to this branch appears to add several input-related lines to `project.godot`. Switching to a different branch (and bringing my changes) reliably reverts it to normal.... Wtfrick is going on here?

GTcreyon:
> So glad this issue has been fixed! This caused problems for me when I was first starting out here.
> 
> On starting the game (at the title screen, like normal), the console gets flooded with this error message:
> 
> ```
> E 0:00:00.689   is_action_just_pressed: The InputMap action "timer_show" doesn't exist.
>   <C++ Error>   Condition "!InputMap::get_singleton()->has_action(p_action)" is true. Returned: false
>   <C++ Source>  main/input_default.cpp:104 @ is_action_just_pressed()
>   <Stack Trace> timer.gd:44 @ _physics_process()
> ```
> 
> Apparently related, I am also unable to map the "show timer" action in the options menu--pressing a new button silently fails and leaves the input unmapped.
> 
> Switching Git to this branch appears to add several input-related lines to `project.godot`. Switching to a different branch (and bringing my changes) reliably reverts it to normal.... Wtfrick is going on here?

I think this was caused by a defective merge commit. It should work fine now. The project wouldn't even load for me. ^^;;

Issue #130, opened 2023-04-01
Spinning into koopa produces error and two shells


nyanpasu64:
**Describe the bug**
When spinning into a wingless koopa, a queue_free warning is printed and two shells are spawned.

**To Reproduce**
- Enter stage 3 of the demo.
- Jump on the parakoopa, then spin into the koopa from the ground on the side.

Results:

- First `_on_HurtboxStrike_body_entered` calls `_hurt_struck(body = player) -> into_shell`.
- Then (on the same frame I think?) `_physics_step()` calls `_hurtbox_check()` which also calls `_hurt_struck`.

`into_shell` is called twice on the same frame (I think), and each call spawns a separate shell and calls `queue_free()` on the koopa. I'm not sure if the koopa is freed before the next `_process` or `_physics_process`.

Sometimes the two shells are spawned at slightly different positions, other times on top of each other (producing a very loud shell sound, and every subsequent kick is loud too because two copies of the shell kick sound play on top of each other).

Because `_hurt_struck` is called twice, it calls `ResidualSFX.new_from_existing` twice, resulting in error:

```
E 0:39:39.579   connect: Signal 'finished' is already connected to given method 'queue_free' in that object.
  <C++ Error>   Method failed. Returning: ERR_INVALID_PARAMETER
  <C++ Source>  core/object.cpp:1480 @ connect()
  <Stack Trace> residual_sfx.gd:42 @ new_from_existing()
                koopa.gd:60 @ _hurt_struck()
                entity_enemy.gd:102 @ _hurtbox_check()
                entity_enemy.gd:95 @ _physics_step()
                entity_enemy_walk.gd:56 @ _physics_step()
                entity.gd:59 @ _physics_process()
```

**Expected behavior**
One koopa shell spawns. I'm not sure which call to `_hurt_struck` should be removed, or if we should make koopa's `_hurt_struck` [idempotent](https://en.wikipedia.org/wiki/Idempotence) so calling it twice does the same thing as calling once (for example we could set a "will be destroyed" flag, and make the method do nothing if we call it again and the flag is already set).

**Screenshots**
![Godot_v3 5 2-stable_win64_yKfQHZMQZB](https://user-images.githubusercontent.com/913957/229302806-132d1110-797b-4729-b77b-b84234002366.png)

**System Information**
 - OS/Browser: Windows 10
 - Game version: master 93835e18640ed009ae033c12dac0cc8c453b347b + `$Damage.set_deferred("monitoring", false)`
[...]

PR #133, opened 2023-04-02
Upgrade ResidualSFX.new_from_existing error handling

Koopa1018:
# Description of changes
Upgrades `ResidualSFX.new_from_existing` to handle more errors, better.

In particular, it now does the following:

- Crashes if the node it was given is not an audio source. (This can be downgraded to an error if needed.)
- Prints an error if the node it was given is an `AudioStreamPlayer3D`. Since the game is 2D, these probably shouldn't be in use anyway.
- Prints an error if the destroy-on-finished-playing signal is connected twice. This situation already errors, but handling it manually lets us give more specific information: what exactly is going wrong, what's been known to cause this before, where to check to try and fix it.

# Issue(s)
Was supposed to deal with #130. It didn't, but it should still be helpful in fixing that bug.

### DIFF
--- a/classes/misc/residual_sfx/residual_sfx.gd
+++ b/classes/misc/residual_sfx/residual_sfx.gd
@@ -49,8 +49,9 @@ static func new_from_existing (
 		# Reparent to the scene root.
 		_move_node_to(sfx, scene_root)
 	
-	# Make the sfx player destroy on playback (if desired)
-	if destroy_on_finished:
+	# Make the sfx player destroy on playback (if desired).
+	# (and if it's not already set that way--Godot complains otherwise).
+	if destroy_on_finished and !sfx.is_connected("finished", sfx, "queue_free"):
 		sfx.connect("finished", sfx, "queue_free")
 	# Lesgo
 	sfx.play()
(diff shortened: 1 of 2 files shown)


classes/misc/residual_sfx/residual_sfx.gd
```diff
@@ -24,10 +24,22 @@ func _init(sound: AudioStream, pos: Vector2):
 # Takes an existing AudioStreamPlayer(2D) and makes it act like a ResidualSFX
 # (outlive its parent, optionally destroy on finish).
 static func new_from_existing (
-	sfx,
+	sfx: Node,
 	scene_root: Node,
 	destroy_on_finished = true
 ):
+	# VALIDATE: Is the passed node an audio source?
+	assert(
```
GTcreyon, line 32: A crash seems excessive. An error should be fine, no? We're not handling saved data, nor would anything particularly serious occur should this be incorrect.

classes/misc/residual_sfx/residual_sfx.gd
```diff
@@ -24,10 +24,22 @@ func _init(sound: AudioStream, pos: Vector2):
...
+		sfx is AudioStreamPlayer or sfx is AudioStreamPlayer2D \
+			or sfx is AudioStreamPlayer3D, # just for completeness
```
GTcreyon, line 34: Please capitalize comments.

GTcreyon (changes requested):
Overall good, but I think demoting to an error rather than a crash seems best.

Koopa1018:
Arright, fixed.

GTcreyon (approved):

PR #139, opened 2023-04-15
Enhance HUD sprites

GTcreyon:
# Description of changes
Replaces old HUD sprites with updated ones. Uses atlases in some cases to reduce load times.


gui/hud/hud_x.png
Koopa1018, line 1: sure you don't want to atlas this with the other HUD elements?

Koopa1018 (approved):
As changes go, this update seems pretty foolproof.

gui/hud/hud_x.png
GTcreyon: Done.

PR #149, opened 2023-04-21
Fix underwater log riding

### LINKED ISSUE #104: Riding falling logs underwater
**Describe the bug**
While standing on a falling log, you can keep standing on it even when underwater (this gets cancelled when you touch a slope)

**To Reproduce**
Step 1: Go to room 1
Step 2: Stand on the very right of the log that falls down (not too far as that will not make you stand on it and put you in the falling state)
Step 3: Wait

**Expected behavior**
What should have happened is that you stopped standing on the log and started swimming

**Screenshots**
image

**System Information**
 - OS/Browser: Michaelsoft Binbows 10
 - Game version: v0.1.6.RC1
 - Other specs: Keyboard

**Additional context**
No

Kuma-Boo:
# Description of changes

As far as I know, there is never a case where the player needs to be registered as "grounded" when entering the water, so this simple one-line fix hopefully won't cause any problems in the future.

# Issue(s)

Closes #104

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -222,6 +222,8 @@ func _on_BackupAngle_body_exited(_body):
 
 var water_areas: int = 0
 func _on_WaterCheck_area_entered(_area):
+	# Disconnect from the ground and start swimming
+	grounded = false
 	swimming = true
 	water_areas += 1
 	# Reset state to normal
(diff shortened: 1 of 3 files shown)


GTcreyon:
I object to the notion that grounded should definitely be false underwater. Mario can stand on the seabed.

Kuma-Boo:
I completely tunnel-visioned on fixing the bug and forgot about that :P

Here's an alternative approach that uses a water-check trigger in log_fall.tscn instead. While I'm a bit unsure about whether this approach is stable at handling entities other than the player, it seems to work alright from my testing.

I think the ideal way to fix this would be to have a list of all the current riders and disconnect them individually as they enter the water, although that seems like a pretty major rewrite.

PR #154, opened 2023-04-21
Fix spin sound not stopping properly

### LINKED ISSUE #101: Diving on the same frame as spinning makes the spin sound for as long as you keep diving
**Describe the bug**
When you spin and dive on the same frame, the spinning sound keeps playing for as long as you are still in the diving animation.

**To Reproduce**
Press X and down arrow at the same time

**Expected behavior**
The spin sound should stop

**Screenshots**
That wouldn't even help

**System Information**
 - OS/Browser: Windows 10
 - Game version: v0.1.6.RC1
 - Other specs: Has motherboard, CPU and GPU 👍

**Additional context**
This also happens with ground pounding, although that one stops when touching the ground so it's less noticeable

Kuma-Boo:
# Description of changes
For some reason, sfx.stop() doesn't work when used on the same frame as sfx.play(). Setting stream_paused doesn't seem to have any issues though, so this fix makes spin_sfx use that instead. 

# Issue(s)

Closes #101

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -1327,9 +1327,10 @@ const STAND_BOX_EXTENTS = Vector2(6, 14.5)
 const DIVE_BOX_POS = Vector2(0, 10)
 const DIVE_BOX_EXTENTS = Vector2(6, 6)
 func switch_state(new_state):
+	# Always pause spin sfx
+	spin_sfx.stream_paused = true
 	# If spin just ended, adjust SFX accordingly.
 	if state == S.SPIN:
-		spin_sfx.stop()
 		if swimming:
 			play_sfx("spin_end", "water")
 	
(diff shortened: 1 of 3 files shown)


Koopa1018 (changes requested):
Just a heads-up: making PRs from your own fork of the codebase makes the branch harder to find in GitHub Desktop. Consider directly branching `master` instead in the future (and if you don't have that permission yet, let Creyon know).

Other than that, mostly seems good. There is a strange bug where opening then closing the debug console makes the spin sound start for no reason. I can't figure out why it happens, so I'll just ping @GTcreyon and hope for the best.

Nice job on the comment formatting, btw :)

Kuma-Boo:
I'll look into it!

GTcreyon (approved):

PR #162, opened 2023-04-21
Make underwater SFX respect player's mute settings

Koopa1018:
# Description of changes
Makes the options menu mute sounds that bypass the underwater muffle effect.

Also, renames the bus that bypasses the underwater muffle effect, to make it a more obviously valid choice when routing sound sources.

# Issue(s)
Fixes bug reported on the Discord server.


Koopa1018:
UPDATE:  Another strange bug seems to have surfaced: after muting sound effects, entering water will make a strange "garbage noise" that sounds like it may be the result of sound effect buildup on the reverb bus.

If music is also muted, this only happens on the first time entering water after starting the game.
If music is enabled, this seems to continue happening faintly whenever water is entered.

GTcreyon (approved):

PR #164, opened 2023-04-22
Fix animation not being set properly when reading signs

GTcreyon:
# Description of changes
Fixes an issue where reading a sign while spinning would cause the neutral animation to play instead.

### DIFF
--- a/classes/interactable/interactable_dialog.gd
+++ b/classes/interactable/interactable_dialog.gd
@@ -26,6 +26,9 @@ func _interact_with(body):
 	body.locked = true
 	body.sign_frames = 1
 	
+	# Initiate the sign-reading animation
+	body.sprite.reading_sign = back_sprite
+	
 	# Disable any active sfx on the player
 	body.start_interaction()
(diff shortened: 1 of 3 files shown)


classes/interactable/interactable_dialog.gd
```diff
@@ -25,8 +25,6 @@ func _interact_with(body):
 	body.vel = Vector2.ZERO
 	body.locked = true
 	body.sign_frames = 1
-	if back_sprite:
-		body.sprite.trigger_anim("back")
```
Kuma-Boo, line 29: Since this is the only place that uses back_sprite, the variable declaration could probably be removed from line 8.

GTcreyon: Actually, no. It has occurred to me that there's a reason this is here. If you talk to a toad, the same code is used, but you're not supposed to use the back sprite there.

GTcreyon: That should be fixed now.

Kuma-Boo (approved):

PR #167, opened 2023-04-22
Fix bus of ResidualSFX

GTcreyon:
# Description of changes
Fixes an issue where ResidualSFXes created through `ResidualSFX.new()` would not be on the SFX bus, and as such would not be muted by Mute SFX.


Koopa1018 (approved):
No troubles encountered. Bug confirmed fixed.

PR #169, opened 2023-04-29
Fix reverb carrying over between scenes

### LINKED ISSUE #168: Reverb glitches out when entering water with SFX muted
**Describe the bug**
In certain cases, entering water when sound effects are muted will make a strange "garbage noise." It sounds like it may be the reverb tail of the "start game" sound. The repro steps require the game start sound to have played while the SFX bus was unmuted, hence the oddly specific repro steps. At least in Godot editor, once it's been done once, repeating the steps will not reproduce the bug again until the game is closed and restarted.

If music is enabled, similar garbage reverb sometimes plays very faintly whenever water is entered. (I suspect it's reverb from previous drum hits in the music?)

**To Reproduce**
1. Open a **fresh** game window **at the title screen,** not inside a level.
2. **DO NOT** mute the SFX bus using the main menu's settings page. (Muting music is fine.)
3. Begin game.
4. Mute the SFX bus using the pause menu's options page.
5. Enter water.

**Expected behavior**
No garbage reverb tails play ever, especially not when SFX are muted.

**Screenshots**
[Audio recording of the glitch](https://drive.google.com/file/d/1N9w7b3W1Y8MP5d0wpI8zYvlzoN3x0ofA/view?usp=sharing)

**System Information**
 - OS/Browser: Windows 10
 - Game version: 0.1.6 Alpha
 - Other specs:

**Additional context**
After thinking about it for a bit, I think I know the root cause.

The reverb bus probably begins the game activated. When the game begins, the player object begins activating and deactivating the bus depending on whether it's in water. Since the player always starts on land, the bus gets deactivated while it's still filled with reverb from the game-start sound, and since it's no longer updating, it never has a chance to "drain."

Consequently, when the player hits the water and the bus reactivates, the stored reverb immediately spills out, bypassing the SFX muting because it comes after the muted buses in the signal chain.

So why does it only happen once per window opening? Because when quitting to title, the reverb bus gets forcibly muted, Consequently, the start-game sound never gets a second chance to fill up the reverb buffer. The reverb bus is instead left filled with audio from the previous session--I've been able to confirm this by unmuting underwater, quitting to title and restarting, then re-muting both music and sound. It's just that audio from the previous session tends to be quiet enough that you can't hear it.

So why doesn't it happen when the SFX are unmuted? Yeah...[about that](https://drive.google.com/file/d/1iY-tJBuDupZNO19AmX0D-HlB0zh3LnWV/view?usp=sharing). (Listen closely!)

AidenBeresford:
# Description of changes
I added a `reset_bus_effect` function to the Singleton class, it resets a chosen bus effect in an audio bus. Additionally, I added a call to this function whenever the player is loaded. This is so any left over reverb is automatically cleared when you enter a new scene, preventing leftover garbage reverb from playing. I also fixed a minor typo.

# Issue(s)
Closes #168

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -178,7 +178,10 @@ onready var feet_area: Area2D = $Feet
 
 
 func _ready():
+	
 	switch_state(S.NEUTRAL) # reset state to avoid short mario glitch
+	
+	Singleton.reset_bus_effect("~Water Verb:", 0) # prevents garbage reverb
 	
 	# If we came from another scene, load our data from that scene.
 	if Singleton.warp_location != null:
@@ -1387,7 +1390,7 @@ func off_ground():
 	cancel_ground = true
 
 
-func recieve_health(amount):
+func receive_health(amount):
 	hp = clamp(hp + amount, 0, 8) # TODO - multi HP
 
 
--- a/classes/global/singleton/singleton.gd
+++ b/classes/global/singleton/singleton.gd
@@ -285,3 +285,10 @@ func save_input_map(input_json):
 	file.open("user://controls.json", File.WRITE)
 	file.store_string(input_json) # minimize the amount of time spent with the file open
 	file.close()
+
+
+func reset_bus_effect(channel, effect_idx):
+	var bus_index = AudioServer.get_bus_index(channel)
+	var bus_effect = AudioServer.get_bus_effect(bus_index, effect_idx)
+	AudioServer.remove_bus_effect(bus_index, 0) # this just gets rid of any unwanted gunk stuck in the bus
+	AudioServer.add_bus_effect(bus_index, bus_effect, 0)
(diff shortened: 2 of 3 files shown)


classes/player/player.gd
```diff
@@ -178,8 +178,11 @@ onready var feet_area: Area2D = $Feet
 
 
 func _ready():
+	
 	switch_state(S.NEUTRAL) # reset state to avoid short mario glitch
 	
+	Singleton.reset_bus_effect("~Water Verb:", 0) # prevents garbage reverb
```
RandomCatDude, line 184: should this be called from the player script? i feel like this adds unnecessary code to the player and could be made more robust by calling from somewhere else
probably somewhere that'll directly link it to scene transitions

GTcreyon: That would probably help, but I would relegate that to its own PR. I think this is fine for 0.1.6.

classes/player/player.gd
```diff
@@ -1387,7 +1390,7 @@ func off_ground():
 	cancel_ground = true
 
 
-func recieve_health(amount):
+func receive_health(amount):
```
GTcreyon, line 1393: Please try to limit small changes like this to their own PR. This almost broke a different script, since it left a function call hanging without its respective definition.

It's not a huge issue, just a word of advice. ^^;;

GTcreyon (approved):

PR #170, opened 2023-04-29
Change speed of initial dive animation

GTcreyon:
# Description of changes
Doubles the speed of the initial dive animation to 10FPS instead of 5FPS.


Kuma-Boo (approved):
Looks good to me!

PR #176, opened 2023-05-30
Add warp paintings

Koopa1018:
# Description of changes
Adds enterable warp paintings, complete with a fancy entry animation.

Under the hood, paintings are based on the same class as doors and warp pipes. Press "enter" button -> play entry animation -> go to new scene, or elsewhere in the current scene.

The entry animation is quite involved, featuring a shader with a painting-ripple effect, as well as an entirely new type of screen transition (see `white_warp.gd`). The specifics of the animation depend on if the painting leads to a new scene:

- If going to a new scene, the animation is several seconds long. When the player hits the painting, the UI disappears, and the camera zooms in on the rippling painting until it fills the screen. At that point, the painting appears to "burn through" to white, and the screen then fades into the next scene.
- If going somewhere in the current scene, the animation only lasts one second. The painting ripple settles down quickly, and none of the other effects happen at all before the player reappears in their warp destination. (The way transitions currently work means there's not even a standard `WindowWarp` transition.)

# Before review:
- [ ] It needs to be possible to change the position of the trigger area relative to the painting.
- [ ] After the 0.1.6 update, Mario needs a new jumping-in animation.
- [ ] Sound design. (Though this can theoretically be added later, in a separate PR.)
- [ ] Would be lovely to decouple choice of animation from whether the scene is changing. This would open the door for much more varied entry animations, e.g. the painting plays an animation before the camera zooms in.
- [ ] Probably some others.


GTcreyon:
How's the progress on this? Is a new jump-in animation needed? If so, I'll need to notify the art team.

Koopa1018:
Stopped work because the 4.1 upgrade broke interactables which respond to `"up"`; see #235. The workaround I used for pipes doesn't seem to work for paintings, which means I might be waiting on #244 before I can make the necessary bugfixes.

Jump-in animation needed for sure, but due to the above, priority is debatable.

PR #181, opened 2023-06-11
Refactor terrain system

Koopa1018:
# Description of changes
Refactors the entire terrain polygon system to be more understandable, more DRY, and generally easier to work with in the future.

Specifically: lots of things were renamed to remove ambiguity and confusion, including functions, variables, and scripts; lots of code (mostly in `terrain_border.gd`) has been deduplicated into shared functions; and of course, nearly everything has been commented and type-hinted.

# Issue(s)
N/A

PR #182, opened 2023-06-11
Background terrain support

Koopa1018:
# Description of changes
Modifies terrain system to support background terrain layers.

***NOTE:** This PR is branched from #181. New changes start on May 30th.*

To make background terrain, two features are required: disable-able collision, and changeable Z-order. Collision was dead simple to connect to a checkbox, so the bulk of the changes here went into Z-sorting.

See, the system was designed to render the top border, then bottom, then sides, and use Z-sorting to put them at the right depth relative to each other. The problem was, this meant a terrain object was on no less than three Z-layers (body, sides, top+bottom). If you wanted to put a terrain in the background, it had to move no fewer than 3 steps at a time in order to stay together--which seems both inconvenient and counterintuitive!

With that in mind, I redesigned the system so the draw steps could be reordered in code (surprisingly non-trivial!), then set it up to draw the different borders from back to front--sorting via draw order rather than Z-order. This means entire terrain objects now only take up one Z-layer, which means they can step backwards and forwards easily (and fit between other layers 😉).

# Issue(s)
N/A - wanted to get this workable so I know how to make my grass background look its best.

### DIFF
--- a/classes/solid/terrain/terrain_border.gd
+++ b/classes/solid/terrain/terrain_border.gd
@@ -41,7 +41,7 @@ func add_full(poly: PoolVector2Array):
 	var type_ids: Dictionary = resolve_edge_types(root.edge_types, poly)
 	var latest_index = 0
 	
-	# Draw each edge type in draw order--first sides, then bottom, finally top.
+	# Draw each edge type from back to front: sides, bottom, finally top.
 	# Begin with sides.
 	latest_index = 0
 	# Iterate the polygon until all chains of top-edge have been found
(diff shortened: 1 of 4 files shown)


jaschutte:
Due to the branch this was based on now having merged with main, could you rebase/merge this branch from main again? Hopefully that resolves some of the merge conflicts.

Small nitpick: For in the future, if you make a branch based of another branch, please make a pull request trying to merge into that original branch. Not into main. That prevents conflicts like these.

I'll review it after you thought this branch up to date with main!

Koopa1018:
Huh! I've tried that before and I thought it wasn't possible. Looks like I just didn't get far enough somehow.

Koopa1018:
@jaschutte Merge has been accomplished.

jaschutte (approved):

jaschutte:
Looks great!

PR #184, opened 2023-06-11
Refactor backgrounds

Koopa1018:
# Description of changes
Unites all background scripts into one multi-purpose `bg_plane.gd` script. This will make it easier to author new backgrounds.

`bg_plane.gd` extends `TextureRect`, as the old scripts did. It is designed for canvas objects, and is thus not useful for making terrain or other such objects move with parallax--that will have to be implemented separately.

Apart from what it inherits from `TextureRect`, it has three features: infinite looping, parallax factor, and autoscroll.

Infinite looping works the same way as it did in `hills.gd`, but it's been extended to work on both X and Y axes. The effect can be enabled/disabled per axis to make infinite horizontal/vertical strips instead of full-screen planes--but it can also be enabled/disabled per side of the object, meaning it's theoretically possible to have e.g. an object that extends infinitely upwards, but ends if you go down far enough. This feature hasn't yet been tested much, and in fact I'm not sure there's much point to it, but here it is. (It'll be much more useful in conjunction with `ColorRect`s--extending a plane with an infinite expanse of color. That can't happen yet, though.)

Parallax factor is how much a plane moves relative to the camera (on each axis). A value of 0 means the plane is fixed in screen space, while a value of 1 means the plane moves at the same speed as an object in the foreground. Values higher than 1 are also possible, and will make the plane pass the camera faster than the level objects; this is useful for putting things in the extreme foreground, though it's unclear whether background planes will even be needed in that context.

Autoscroll is fairly self explanatory, and can also be set per axis. The clouds in the current background use this to appear to drift by.

The currently-existing background has been upgraded to use these new scripts. It doesn't look exactly the same as it used to; I'd fix that, but it's getting replaced soon anyway.

In the future, I would like to add some kind of "periodic" looping, which doesn't stretch the plane out infinitely but rather produces infinite spaced-out copies of it. For the indoorsy areas of the game, this could be useful for creating rows of pillars, and in conjunction with autoscrolling, it could be used to make flotsam and jetsam floating down e.g. a river.... But I haven't added that yet, and since it's not yet super pressing, I'll leave it for the future. (May be best to save that for the traditional-sprite parallax, come to think of it.)

# Issue(s)
Need these tools to implement new backgrounds.


backgrounds/bg_plane.gd
```diff
@@ -0,0 +1,96 @@
...
+			margin_bottom = max(screen_size.y, margin_top)
+		else:
+			margin_bottom = margin_top + start_rect_size.y
+			
+		# Counteract camera zoom so BG stays same size onscreen.
+		margin_left *= rect_scale.x
+		margin_top *= rect_scale.x
+
+
+func vec2lerp(from: Vector2, to: Vector2, fac: Vector2):
```
jaschutte, line 92: This is a standard feature, `lerp()` can take in any lerp-erable value, which includes Vector2s

jaschutte (dismissed):

GTcreyon:
This PR has a somewhat disjointed commit history. Why does it include a merge from the 4.1 branch? What *happened* here? o-o

I'm not quite sure how to resolve this issue other than ditching this PR and reimplementing the intended changes, but if anyone has any guidance, feel free to point it out.

Koopa1018:
The 4.1 branch merge was intended to future-proof it. I consider it out of scope of the 4.1 branch (it's not fixing upgrade bugs, after all), but on the other hand, merging it into `master` seems unwise before a big change like 4.1 hits every file in the repo (and causes merge conflicts _outside_ the scope of the relevant PRs).

GTcreyon:
Is this PR still relevant, given that we're redoing the background art? Or, perhaps that should be an extension of this PR?

Koopa1018:
This is a prereq for the background redo. It does need some reconsideration, given that I didn't know `ParallaxBackground` existed when I did the refactor 😅

The new background actually already has its own branch, `asset-new-grass-bg`. It's based on this one. Not sure I ever updated it for 4.1....

Koopa1018:
Refactored PR to use the Godot-native `ParallaxLayer` and `ParallaxBackground` nodes. This can do everything that the original version can do except for one-sided infinite strips (which can be implemented in a different way if they're ever needed), and should theoretically be both easier to maintain and faster.

GTcreyon (approved):
I've always had a prejudice against `ParallaxBackground`. I don't recall why. But I can't argue with results, so I'm happy to shed my prejudices and go ahead with the merge. Good work!

PR #186, opened 2023-06-30
Make LD load again by cutting new item icons

Koopa1018:
# Description of changes
Apparently ever since the 0.1.6 graphics overhaul, opening the Level Designer has crashed the game (editor)/caused catastrophic breakage (web). This PR fixes that.

The problem was, the LD was using loose frames of animation as icons for the item menu. When the loose frames were replaced with atlases, the paths to the loose frames weren't updated--nor could they be, since the frames were now part of large atlases.

To fix this, I cut Atlas Textures (quicker to make than PNGs) of all the offending objects--coins in all colors, Cheep Cheeps, and Silver Shines (which currently use a HUD widget for their icon)--and updated the icon loader to support `.tres` Texture resources in addition to .png files. (Any Texture, not just Atlas Textures--seems silly to limit ourselves :P)

# Issue(s)
N/A - just hard to work on a broken LD ;)


Koopa1018:
Briefly restored to merge into 4.1 branch.

PR #187, opened 2023-06-30
Fix garbage reverb thoroughly

Koopa1018:
# Description of changes
I've caught #168 cropping up here and there still--it's a lot subtler, but on rare occasions, enough can happen immediately before jumping out of water that the leftover reverb tail is audible when it spills over into the next body of water (try superswimming out of water immediately after pounding into it).

To fix this, I moved the reverb reset from `_ready()` to `_on_WaterCheck_area_entered()`. This way, the reverb buffer is forcibly cleared every single time Mario hits water, meaning there's no possible way reverb from last time could spill over.

# Issue(s)
Closes some loose ends on #168.

### DIFF
--- a/classes/global/singleton/singleton.gd
+++ b/classes/global/singleton/singleton.gd
@@ -288,7 +288,10 @@ func save_input_map(input_json):
 
 
 func reset_bus_effect(channel, effect_idx):
-	var bus_index = AudioServer.get_bus_index(channel)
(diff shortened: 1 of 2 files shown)


jaschutte (approved):

Koopa1018:
Briefly restored to merge into 4.1.

PR #189, opened 2023-07-12
Feature command autocompletion

jaschutte:
# Description of changes
The command line is an amazing feature, it allows us to quickly add and test features without replaying it. And for players, it is fun to mess around with. However, currently this feature has no way to know in game which commands are valid and which ones are not. This PR hopes to improve that by providing a list of all commands and argument possibilities. This list can also be used to automatically complete commands and arguments as well by simply pressing enter.

Besides that, it also fixed the command bar not closing.

# Issue(s)
Closes N/A


Kuma-Boo (changes requested):
Looks mostly good! I never knew what the commands were, so this will definitely help a lot.

I feel like it'd be nice if some of the single letter commands had a more descriptive option though (i.e. having both "w" and "warp" command instead of just "w").

Also, this is from before this pr and might need to be a separate issue, but forgetting to pass a second argument results in a crash for any multi-argument command other than "w".

Koopa1018:
Making a loooot of changes here. Best to make multiple PRs so we can give them all enough attention on their own.

jaschutte:
> Making a loooot of changes here. Best to make multiple PRs so we can give them all enough attention on their own.

I mean yeah :sweat_smile: , it's adding an entirely new feature, these tend to be bigger PRs than code re-writes or bug fixes. I can split this up but then we'd have individual features which don't make _that_ much sense on their own. 

However, I should have made issues for this before hand.

If the code looks good, I would like to have this merged as it has been open for quite some time rn.

classes/console/console.gd
```diff
@@ -236,12 +236,16 @@ func enable(enabled):
 	if visible:
 		input_line.grab_focus()
 		input_line.text = ""
+		_on_Input_text_changed("")
```
GTcreyon, line 239: Calling callbacks outside of signals seems like bad practice - is there no other way this could be dealt with?

GTcreyon (changes requested):
Works nicely, just an issue with a line.

classes/console/console.gd
jaschutte: I mean I could move the code inside of the callback into a seperate function and then invoke that from the callback and here.

But that just seems more effort than it seems worth, and the callback is relating on input of the text field changing -which is exactly what is happening here. That's why I invoked it, as there's some logic which needs to run when the text changes, so it all just feels appropriate.

GTcreyon: Hm. Maybe do it by emitting the text_changed signal instead, so it's working as expected for the node and we don't get any unexpected behavior?

jaschutte: Didn't know that was a thing! If you could make that change, much appreciated (not on pc rn)

GTcreyon:
My change was reviewed by @jaschutte. Merging.

PR #203, opened 2023-07-23
Fix main menu item description misalignment

GTcreyon:
# Description of changes
The descriptions on the main menu had misconfigured control alignments. This fixes that.


GTcreyon:
Oops, that's not right. Shouldn't be merging into `master`.

GTcreyon:
Fixed.

Koopa1018 (changes requested):
Nice, I thought centering would be the right call here. Vertical centering seems off, though: the text runs up against the bottom of the box.

GTcreyon:
> Nice, I thought centering would be the right call here. Vertical centering seems off, though: the text runs up against the bottom of the box.

That's normal - its always been like that.

Koopa1018 (approved):

PR #204, opened 2023-07-23
Fix scaling on title screen

GTcreyon:
# Description of changes
Fixes a few issues causing elements of the title screen not to scale correctly.


Koopa1018 (approved):
Yep, looks fixed!

PR #205, opened 2023-07-25
Fix all spots which give int division warnings

Koopa1018:
# Description of changes
Fix all remaining spots which give int-division warnings.

Spots commented with `# warning-ignore:integer_division` were upgraded to use the proper Godot 4 version of that command (`@warning_ignore("integer_division")`).

I'm assuming other erroring divisions were expected to divide as float, so any such spots have been upgraded with the required casts to make that happen.

A lot of these warnings came from spots that were using `get_window().size`. Many of them called this function multiple times per function, or even per line, and that seems a bit stinky...so I rewrote them to call once, convert to a float vector, then reuse the converted vector. Also saves horizontal real estate in the scripts.

### DIFF
--- a/classes/entity/passive/goonie/goonie.gd
+++ b/classes/entity/passive/goonie/goonie.gd
@@ -39,7 +39,7 @@ func _ready() -> void:
 	if not disabled:
 		sprite.play()
 	
-	# warning-ignore:integer_division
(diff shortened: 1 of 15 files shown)


jaschutte:
Checked this some time ago, but forgot to do anything with it hehe

PR #215, opened 2023-08-05
Fix fungus stem tiling

### LINKED ISSUE #206: Fungus platform stems don't tile properly
**Describe the bug**
The spotted texture on the fungus platforms' stems does not tile across the length of the stem. It appears once below the head, but the rest of the stem is filled with ONLY the bottom row of the texture.

**To Reproduce**
Reach any fungus platform in T1.

**Expected behavior**
The stem texture repeats along the length, as it did in Godot 3.5.

**Screenshots**
image

**System Information**
 - OS/Browser: Windows 10 Godot Editor
 - Game version: 4.1 Branch
 - Other specs: GPU is RTX 2070 Super, if that makes a difference

**Additional context**
I checked the stem scene, and the texture mode is set to "Tile" as you would expect....

Kuma-Boo:
# Description of changes

Force fungus stems to repeat their textures.

# Issue(s)

Closes #206.

### DIFF
--- a/classes/solid/fungus_platform/fungus_stem.tscn
+++ b/classes/solid/fungus_platform/fungus_stem.tscn
@@ -5,6 +5,7 @@
 
 [node name="FungusStem" type="Line2D"]
 z_index = -2
+texture_repeat = 2
 position = Vector2(6, 10)
 points = PackedVector2Array(-6, -3, -6, 6)
 width = 8.0


Koopa1018 (approved):
Wow I just. Literally did not realize this property existed. Kudos to you for digging it up.

PR #219, opened 2023-08-08
Update Toad.gd to allow for easier editing

Kuma-Boo:
# Description of changes

Enables a live preview of toad's color palette in the editor.

### DIFF
--- a/classes/interactable/npc/toad/toad.gd
+++ b/classes/interactable/npc/toad/toad.gd
@@ -1,3 +1,4 @@
+@tool
 class_name Toad
 extends InteractableDialog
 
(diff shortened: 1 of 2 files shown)


classes/interactable/npc/toad/toad.gd
```diff
@@ -41,29 +42,27 @@ const skin_presets = [
...
 func set_skin(new_skin):
 	for i in range(4):
 		material.set_shader_parameter("color" + str(i + 2), skin_presets[new_skin][i])
 	skin = new_skin
-	temp_skin = new_skin
 
 
 func _ready():
+	set_spot(spot)
+	set_skin(skin)
```
Koopa1018, line 61: These two function calls aren't necessary--the variables' `set` clauses run them just fine.

Koopa1018 (approved):
Looking good. Always wondered what these alternate palettes looked like.

classes/interactable/npc/toad/toad.gd
Kuma-Boo: Those are there so Toads initialize with the correct color when instanced into the editor tree. Without it, you'd need to manually switch the spot and skin properties for the palette to load properly.

Koopa1018: Ah. That makes sense. Would suggest mentioning that in a code comment, then (so future readers know not to refactor it out).

PR #220, opened 2023-08-08
Fix non-integer-division code warnings

Koopa1018:
# Description of changes
Fixes all remaining runtime warnings (that I could find).

I'm not certain about the use of `@warning_ignore("shadowed_variable")`. The affected functions' readability would take a hit if I renamed the variables instead, I think, but it still feels icky. Suggestions? (relevant scripts are `player`, `character_sprite`, `terrain_border`, and `pause_button`)

# Issue(s)
N/A


classes/player/player.gd
```diff
@@ -1476,6 +1476,7 @@ func resist(val, sub, div): # ripped from source
 	return val * FPS_MOD
 
 
+@warning_ignore("shadowed_variable")
 func set_rotation_origin(facing_direction: int, origin: Vector2):
```
Kuma-Boo, line 1480: We could avoid shadowing the variable by changing "facing_direction" to "direction" and multiplying origin.x by it.
i.e.
```suggestion
func set_rotation_origin(direction: int, origin: Vector2):
	# Flip origin, if appropriate.
	origin.x *= direction
	sprite.offset = origin
	sprite.position = -origin
```

classes/solid/terrain/terrain_border.gd
```diff
@@ -226,6 +226,7 @@ func _add_inbetween_segment(areas, start: Vector2, end: Vector2, circumcenter: V
 	})
 
 
+@warning_ignore("shadowed_variable_base_class")
 func generate_polygons(lines: Array, texture: Texture2D, z_index: int):
```
Kuma-Boo, line 230: Can be renamed to "z_order" to match generate_polygons_top()

gui/pause/buttons/pause_button.gd
```diff
@@ -53,8 +53,9 @@ func resize():
 	stars.polygon[2].x = size.x - 1
 
 
-func _on_Button_toggled(button_pressed):
-	if button_pressed:
+@warning_ignore("shadowed_variable_base_class")
+func _on_Button_toggled(is_pressed: bool):
```
Kuma-Boo, line 57: This should probably follow Godot conventions. The passed parameter can also just be safely discarded.
```suggestion
func _on_button_toggled(_button_pressed: bool):
```

Kuma-Boo (changes requested):
Looks mostly good! Though I don't really like the idea of ignoring shadowed variables. I feel like it's even less readable than renamed variables honestly.

Also, the variable in character sprite "frame" (at _is_footstep_frame()) can probably also be safely removed.

gui/pause/buttons/pause_button.gd
Koopa1018: The parameter's checked in the very next line???

And yeah, I know the name's unconventional. Auto-generated signal functions are just Like That. We might want to do a project-wide update-and-reconnect session, because this sure ain't the only one. (Or leave it, since it's instantly recognizable as A Signal Function.)

classes/solid/terrain/terrain_border.gd
Koopa1018: Ooh, good thought.

classes/player/player.gd
Koopa1018: Hmm. Just `direction` is objectively more ambiguous, though.... Maybe `facing_dir`?

Good thought about multiplying `origin.x` though. Definitely a more elegant way to do it!

Koopa1018:
That variable gets used in the very last line of the function. It's...kind of the whole point of the function.

gui/pause/buttons/pause_button.gd
Kuma-Boo: `is_pressed` is the getter function for the `button_pressed` property in the base Button class. Since this property is updated before the signal is called, the function still works properly even when discarding the signal parameter. Godot probably only passes the parameter in case the signal is connected to a different node, so I don't think we need it here.

Thought leaving the signal names alone is probably a good idea for now.

Kuma-Boo:
> That variable gets used in the very last line of the function. It's...kind of the whole point of the function.

Don't know what I was thinking here, thank goodness I'm not programming this by myself. :P

gui/pause/buttons/pause_button.gd
Koopa1018: Fair point, but I always prefer to use the most local copy of the data available. Helps keep functions as pure as possible.

Koopa1018:
As I check, though, you were right. It would've pulled from the class's `frame` variable, just like in the `is_pressed` situation. Not ideal for function purity, but it would've worked just the same.

PR #221, opened 2023-08-08
Fix editor errors caused by toolscripts

Koopa1018:
# Description of changes
In the editor, a number of toolscripts throw errors on scene load, due to trying to `@onready` get nodes that don't exist yet. This PR fixes that using null checks, later initialization, and in `camera_area`'s case, just un-toolscripting things that don't need it.

# Issue(s)
N/A


Kuma-Boo (approved):
Looks good!

PR #222, opened 2023-08-09
Fix life hud flashing

Kuma-Boo:
# Description of changes

Very simple fix that prevents the life hud's coin meter from flashing when changing scenes.


Koopa1018 (approved):
Sorry I've been so long in reviewing. Looks like it's fixed!

Issue #118, opened 2023-03-26
Spin attack hitbox is uncomfortably small


Koopa1018:
**Describe the bug**
Spinning hits enemies from an uncomfortably small range away.

The player's normal hitbox is slightly narrower than the player sprite when standing.  During a spin, the sprite's silhouette becomes several pixels wider, but the hitbox remains narrow, leading to cognitive dissonance when Mario's arms visibly touch the enemy, but nothing happens.

This is in contrast with the Flash version, wherein the spin attack hitbox was almost excessively generous.

**To Reproduce**
Our established muscle memory may obscure the problem. Recommend trying these steps on a fresh individual.
1. Allow a goomba to charge you.
2. Time a spin attack using only visual information (the default for many players)

**Expected behavior**
Spin attack range matches or exceeds what the player can see.

**Screenshots**
Arm visibly hitting the box, yet the box is unbroken.
![image](https://user-images.githubusercontent.com/29995463/227748490-b78bf710-c971-4a47-b03c-57a7cb4120ce.png)

**System Information**
 - OS/Browser: Win10
 - Game version: 0.1.6 alpha
 - Other specs: N/A

**Additional context**
Would recommend extending the hitbox past the visible boundaries slightly. as an affordance to players. (Though maybe not quite as much as the Flash version--too much of a good thing....)

Perhaps don't fix until the new spin sprites drop.

Kuma-Boo:
Maybe instead of having two separate hitbox sizes, we can just have a timer that keeps track of how long the player's been spinning for.

That way, environmental objects like this will always consistently break, while enemies can check the timer to see if they should take damage or not.

Koopa1018:
Nah, it's not a box-specific problem. (Boxes were just easy to screenshot acting weird, since they don't move.) Affecting enemies is what I'm going for here.

Koopa1018:
Notes on Flash version's spin, copied from Kuma's comment in #223:

> After making some comparisons with the original flash version, The Flash version has an absurdly large hitbox, but seems to only be active for about 12 frames. Redux's spin in comparison has a super long active window of 30 frames.
> 
> Not really a big deal as we can always nerf it later and it's pretty unrelated to this pr, but I just thought I should document it.

GTcreyon:
> The Flash version has an absurdly large hitbox, but seems to only be active for about 12 frames. Redux's spin in comparison has a super long active window of 30 frames.

[...]

PR #223, opened 2023-08-09
Fix enemy hitbox priorities

### LINKED ISSUE #128: Spinning into enemy at high speed randomly damages player instead of enemy
**Describe the bug**
When I spin and enter an enemy (eg. goomba) at a high speed, sometimes Mario takes damage instead of the Goomba.

**To Reproduce**
- Enter room `tutorial_1_4`, run right, and spin before hitting the goomba.

Unsure if the same bug causes taking damage from Cheep Cheeps during the damaging part of a player spin, or another issue.

## Adding logs

How do Mario and goombas interact? In `entity_enemy.gd`, the goomba sets up collision checks with the player:

```
func _connect_signals():
	._connect_signals()
	_connect_node_signal_if_exists(hurtbox_stomp, "area_entered", self, "_on_HurtboxStomp_area_entered")
	# Take damage from Mario
	_connect_node_signal_if_exists(hurtbox_strike, "body_entered", self, "_on_HurtboxStrike_body_entered")
	# Attack Mario
	_connect_node_signal_if_exists(hitbox, "body_entered", self, "_on_Hitbox_body_entered")
```

To figure out the order that Mario and Goomba's physics ticks, and their collision handlers, execute, I [added debug prints to these methods](https://github.com/nyanpasu64/SM63Redux/tree/debug-spin-hurt).

Every physics tick, Mario's movement runs before the goomba's movement. How do I know?

- After unpausing, the first print is `player.player_physics()` before `goomba._physics_step()`.
- Player-enemy collision checks always occur after the goomba moves, but before Mario moves.
	- When Mario and the enemy run into each other from the side, both `_on_HurtboxStrike_body_entered` and `_on_Hitbox_body_entered` always occur on the same (frame? physics tick?), because the goomba's attack hitbox and strike hurtbox are the same width. This may not necessarily be the case for vertical movement.
- If I place a breakpoint on each of these methods, camera movement occurs after the goomba moves.
	- Interestingly, when I continue the game past Mario's physics tick but break the game when the goomba's physics tick begins, Mario moves relative to the goomba and camera!
	- This is bizarre, apparently GDScript breakpoints freeze the physics "thread" but not the rendering code, which plots an *inconsistent intermediate state* (mid-physics-tick) of the game on-screen?!

Oddly, collision checks apparently occur the frame *after* the player and enemy move?

- One time I unpaused the game, and the first breakpoint to hit was the *collision checks* resulting from overlapping player/enemy positions. This means that the player and goomba's last physics step *before* pausing caused them to overlap (setting Mario up to take damage), but the actual collision damage callbacks were called *after unpausing?!*
- Similarly, I've managed to get the translucent pause menu to disappear *after* Mario and the goomba moved, but before they processed their collision from that movement.

## Testing the bug

As to why the bug occurs... when I spin into a goomba and kill it, the goomba's `_on_HurtboxStrike_body_entered` method is called before its `_on_Hitbox_body_entered` method on the same (frame? physics tick? I don't know!):

```
player.player_physics()
goomba._physics_step()
player.player_physics()
goomba._physics_step()
entity_enemy#_on_HurtboxStrike_body_entered()
	player attacking = True
entity_enemy#_on_Hitbox_body_entered()
	enemy attacking = False
player.player_physics()
goomba._physics_step()
```

When the goomba damages Mario, the goomba's `_on_Hitbox_body_entered` (attack) method is called first, before Mario's collision with the goomba's hurtbox is processed:

```
player.player_physics()
goomba._physics_step()
player.player_physics()
goomba._physics_step()
entity_enemy#_on_Hitbox_body_entered()
	enemy attacking = True
entity_enemy#_on_HurtboxStrike_body_entered()
	player attacking = False
player.player_physics()
goomba._physics_step()
```

When Mario is spinning and moving slowly compared to the goomba, both methods are run (not just Mario hitting the goomba), but the hurtbox method is usually called first, and Mario usually damages the goomba successfully.

## Speculation

One possibility is that if Mario moves into the goomba and collides into the goomba's "position the previous tick", then `_on_Hitbox_body_entered` (enemy attacks Mario) gets called first. And if the goomba moves from "not touching Mario's new position" to "colliding with Mario" at the end of the physics tick, then `_on_HurtboxStrike_body_entered` (Mario attacks enemy) gets called first.

But I don't know how the order of physics movements could affect the order of callbacks, since I hear callback order is related to node tree order, both connections are made in the same direction, both `hurtbox_strike` and `hitbox` belong to the goomba and don't change in order (I think? could be wrong), and the player-goomba physics move the goomba as a full unit, rather than its hitboxes individually.

- But perhaps moving a goomba moves/processes its hitboxes in a different order, than moving a player into the goomba's hitbox?

**Expected behavior**
Spinning into an enemy which can be stunned or defeated by a spin, always damages the enemy without the player taking damage.

I don't know how this could possibly be implemented given Godot's current callback system, unless you can control the order of callbacks, or buffer all player-enemy collision events and process them once a physics tick in a consistent manner (independent of the order of collision events).

**Screenshots**

https://user-images.githubusercontent.com/913957/228808593-a5b17e7f-02a3-4ddc-b976-d249884e720e.mp4

**System Information**
 - OS/Browser: Windows 10 Ameliorated
 - Game version: Git master 5ff30053014e81fd8e96c94b18e51b3e70f6adc4
 - Other specs:

**Additional context**
Possibly related:

- #118

Kuma-Boo:
# Description of changes

Actual changes:
- Enemy entities use a new "deferred" connect mode to ensure their signals are always processed after the player.
- While fixing the parakoopa's hitboxes, I also refactored its script a bit.

Some extra things to note:
- While spinning enemies is more consistent, the timing feels way too generous now. Will open a proposal to fix this later.
- The transitions between parakoopas, koopas and shells are still super jank. Will open a proposal to clean them up later.

# Issue(s)

Closes #128
Fixes the first reproduction scenario of #118

### DIFF
--- a/classes/entity/enemy/koopa/parakoopa.gd
+++ b/classes/entity/enemy/koopa/parakoopa.gd
@@ -22,13 +22,15 @@ const color_presets = [
 
 @export var disabled = false: set = set_disabled
 @export var mirror = false
-@export var color = 0: set = set_color
+@export var color: Koopa.ShellColor = Koopa.ShellColor.GREEN: set = set_color
 
 
 func set_color(new_color):
 	for i in range(3):
 		material.set_shader_parameter("color" + str(i), color_presets[new_color][i])
 	color = new_color
+	koopa.color = color
+	shell.color = color
 
 
 func _ready():
@@ -50,27 +52,21 @@ func _exit_tree():
 func _on_TopCollision_body_entered(body):
 	if body.is_spinning():
 		spawn_shell(body)
-	else:
-		if body.vel.y > -2:
-			$Kick.play()
-			koopa.position = Vector2(position.x, body.position.y + 33)
-			koopa.vel.y = body.vel.y
-			koopa.mirror = flip_h
-			koopa.color = color
-			body.vel.y = -5.5
-			body.vel.x *= 1.2
-			get_parent().call_deferred("add_child", koopa)
-			$TopCollision.set_deferred("monitoring", false)
-			$Damage.set_deferred("monitoring", false)
-			set_deferred("visible", false)
-			visible = false
-
-
-func _on_Kick_finished():
-	queue_free()
+	elif body.vel.y > -2:
+		change_state()
+		get_parent().call_deferred("add_child", koopa)
+		koopa.position = Vector2(position.x, body.position.y + 33)
+		koopa.vel.y = body.vel.y
+		koopa.mirror = flip_h
+		body.vel.y = -5
+		body.vel.x *= 1.2
 
 
 func _on_Damage_body_entered(body):
+	# Parakoopa was already defeated, return early
+	if !visible:
+		return
+	
 	if !body.is_diving(true):
 		if body.is_spinning():
 			spawn_shell(body)
@@ -82,26 +78,26 @@ func _on_Damage_body_entered(body):
 
 
 func spawn_shell(body):
-	$Kick.play()
+	change_state()
 	body.vel.y = -5
 	get_parent().call_deferred("add_child", shell)
 	shell.position = position + Vector2(0, 7.5)
-	shell.color = color
 	if body.global_position.x < global_position.x:
 		shell.vel.x = 5
 	else:
 		shell.vel.x = -5
-	$TopCollision.set_deferred("monitoring", false)
-	$Damage.set_deferred("monitoring", false)
-	set_deferred("visible", false)
+
+
+# Called whenever the parakoopa loses its wings
+func change_state():
+	$Kick.play()
+	top_collision.set_deferred("monitoring", false)
+	hurtbox.set_deferred("monitoring", false)
+	visible = false
 
 
 func set_disabled(val):
 	disabled = val
-	if hurtbox == null:
-		hurtbox = $Damage
-	if top_collision == null:
-		top_collision = $TopCollision
 	hurtbox.monitoring = !val
 	top_collision.monitoring = !val
 	if !Engine.is_editor_hint():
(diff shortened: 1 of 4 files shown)


classes/entity/enemy/koopa/parakoopa.gd
```diff
@@ -82,26 +78,26 @@ func _on_Damage_body_entered(body):
...
 		shell.vel.x = 5
 	else:
 		shell.vel.x = -5
-	$TopCollision.set_deferred("monitoring", false)
-	$Damage.set_deferred("monitoring", false)
-	set_deferred("visible", false)
+
+
+# Called whenever the parakoopa loses its wings
+func change_state():
```
Koopa1018, line 92: Would suggest calling this `defeat()` instead. `change_state()` implies that there's more than one state it could go to.

Koopa1018 (approved):
Niiiiiice. Spinning finally works like it should!

(Spin timing seems perfectly fine btw. You wanna see _generous_, you should check out the Flash version 😝)

Kuma-Boo:
After making some comparisons with the original flash version, The Flash version has an absurdly large hitbox, but seems to only be active for about 12 frames. Redux's spin in comparison has a super long active window of 30 frames.

Not really a big deal as we can always nerf it later and it's pretty unrelated to this pr, but I just thought I should document it.

Koopa1018:
Hmm. That duration change is a good thing to note.

Koopa1018:
(come to think of it I suppose I was talking about the giant hitbox before, not the timing...my bad 🤦)

Koopa1018:
(whoa. just checked the timing out and dang you weren't kidding--that window is BIG.)

PR #226, opened 2023-08-11
Fix main menu texture repeats

### LINKED ISSUE #225: Main menu buttons' shine lines display wrong
**Describe the bug**
The main menu's buttons ("Story Mode", "Level Designer," etc.) normally have diagonal white shine lines across their lower edges. Since the 4.1 update, these shine lines don't appear except at the very left edge of the screen, meaning 2 out of 3 visible buttons' lower edges are just transparent.

Also, the lines which _do_ appear seem to be straight vertical, rather than diagonal.

**To Reproduce**
Enter the main menu.

**Expected behavior**
Shine lines appear on all buttons, and run diagonally right from top to bottom.

**Screenshots**
Left: release v0.1.6. Right: 4.1 branch.
image

**System Information**
 - OS/Browser: Windows 10
 - Game version: 4.1 branch
 - Other specs: 

**Additional Context**
I suspect the left-side problem is a texture wrapping problem--we've certainly had a few of those.
Come to think of it, the not-diagonal problem could also be that (extending pixels past the edge).

Koopa1018:
# Description of changes
Makes all textures inside the main menu repeat by default. This fixes the diagonal shines on the buttons, as well as the tiling on the apparently unused Line2D called "Stripe."

While I'm at it, also renames the diagonal shines to match their script name (which is "sheen," not "shine"--much clearer in a game about shines).

# Issue(s)
Closes #225, both parts of it.

### DIFF
--- a/scenes/menus/title/main_menu/main_menu.tscn
+++ b/scenes/menus/title/main_menu/main_menu.tscn
@@ -222,6 +222,7 @@ animations = [{
 }]
 
 [node name="Menu" type="Control"]
+texture_repeat = 2
 layout_mode = 3
 anchors_preset = 15
 anchor_right = 1.0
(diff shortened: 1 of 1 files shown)


Kuma-Boo (commented):
Looks good!

PR #227, opened 2023-08-14
Increase sign glow size (and other changes that need discussion)

### LINKED ISSUE #224: Glow outlines are too thin
**Describe the bug**
Since the 4.1 update, signs' outlines are only 1px thick instead of however wide they were in 3.5.x.

**To Reproduce**
Approach a sign.

**Expected behavior**
Glow outlines are the same width as 3.5.

**Screenshots**
 Before / After
image

**System Information**
 - OS/Browser: Windows 10
 - Game version: 4.1 branch
 - Other specs:

Kuma-Boo:
# Description of changes
 - Increases sign glow by one pixel.
 - Gave signs their own material.
 - Duplicate materials at runtime instead of in editor

# Extra notes
Currently every object that uses glow has its own local copy of `glow.tres` (`resource_local_to_scene`), so should we ever need to make these kind of changes to an object, we would have to go through every single scene that has said object and manually change it.

Ideally we could use [instance shader uniforms](https://godotengine.org/article/godot-40-gets-global-and-instance-shader-uniforms/), but it doesn't seem to work for canvasitems in 4.1, leaving us with two workarounds:

1. Have glowing objects duplicate their materials at runtime (used in this pr)
2. Brute force it and hope we never have to change anything

# Issue(s)

Closes #224


Koopa1018:
Personally I'm all for the duplicate-at-runtime approach. Currently, having them local to scene generates a LOT of version control noise (their IDs change every time the scene is loaded--a known problem in Godot, hopefully fixed in the future, but not yet).

Koopa1018:
Eh. I kinda doubt this change actually touches any areas outside of signs. Let's get this thing reviewed.

Koopa1018 (approved):
All looks good.

PR #235, opened 2023-09-05
Make pipes usable again

Koopa1018:
# Description of changes
Somewhere along the line, pipes were set to no longer monitor the player's layer. This PR simply fixes that, allowing pipes to work again.

# Issue(s)
N/A

### DIFF
--- a/classes/interactable/pipe/pipe.tscn
+++ b/classes/interactable/pipe/pipe.tscn
@@ -11,7 +11,7 @@ size = Vector2(8.73426, 16)
 size = Vector2(32, 32)
 
 [node name="Pipe" type="Area2D"]
-collision_mask = 0
+collision_mask = 2
 monitorable = false
 script = ExtResource("3")
 sprite = NodePath("Sprite2D")


jaschutte (changes requested):
This does indeed make the pipe intractable again, trying to enter the pipe using the down arrow doesn't work, using Z to ground pound does work.

Although this bug isn't reaaally related to this PR, since this is such a small PR, could you take a look at this bug as well?

Koopa1018:
That's already known. It happens when the pipe comes after Mario in the hierarchy--since execution flows from top to bottom, Mario's dive/crouch code somehow interferes with the pipe's activation. It happens with doors too.

Pre-4.1, I had this fixed by setting pipes' execution order to -1, but since the upgrade, it doesn't let me set it lower than 0 anymore...even via code, it seems like (because I know that worked before). I have no idea how else to go about this...any thoughts?

Either way, putting the pipe before Mario is an effective workaround. 

(And I'd say this bug definitely fits in this PR. Different root causes, but the effect is the same: pipes don't activate when they should.)

jaschutte:
If the root issue is the execution flow, could we not invoke the pipe logic from Mario? Or would that over-complicate the code? I haven't had a good look at the Mario logic code.

But to me, the  best solution would be to check for pipes within Mario before the flag gets cleared. However this would make Mario implement logic technically meant for pipes.

Alternatively we could have like an array of all flags which have been _enabled_ within **this** frame for Mario, then the pipe could simply read from that array. But this could be very bad practice as well as other code would more quickly pull from this array rather than doing it 'properly'.

I'd suggest trying to invoke pipe logic inside of the Mario, that seems like the best solution.

Koopa1018:
Hmmmm. I can actually see that working.

Pipes and doors are a special case of `Interactable`. Would we want all `Interactable`s handled by the player? Or would we want to make pipes and doors a separate class? (They kind of already are--they inherit `InteractableWarp`, which inherits from `Interactable`.)

Either way, that's big enough that we really should make a whole separate PR for it.

GTcreyon:
Merging this now - we'll fix the overall issue later.

PR #239, opened 2023-09-16
Fix Cheep Cheep animations

### LINKED ISSUE #238: Cheep Cheep animations are broken
**Describe the bug**
Looks like one of the Cheep Cheeps' animations is set to loop incorrectly?

**To Reproduce**
Watch a Cheep Cheep as it chases the Player.

**Expected behavior**
Cheep Cheep animations match release 0.1.6.alpha.

**Screenshots**
Animation glitch, need video.

**System Information**
 - OS/Browser: Windows 10, Godot editor
 - Game version: 4.1 branch
 - Other specs:

**Additional context**
Creyon probably knows better what they're supposed to look like....

Koopa1018:
# Description of changes
Sets Cheep Cheeps' transition animations to not loop. Animations fire different signals when they loop versus when they end, and because the code is wired to the end signal, it was never advancing to the loop portion....

# Issue(s)
Closes #238.

### DIFF
--- a/classes/entity/enemy/cheep_cheep/cheep_cheep.tscn
+++ b/classes/entity/enemy/cheep_cheep/cheep_cheep.tscn
@@ -99,7 +99,7 @@ animations = [{
 "duration": 1.0,
 "texture": SubResource("29")
 }],
-"loop": true,
+"loop": false,
 "name": &"calmdown",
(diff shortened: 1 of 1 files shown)


Kuma-Boo (approved):
Looks good! 👍

PR #241, opened 2023-09-16
Fix ground pounding moving platforms

Kuma-Boo:
# Description of changes

~~Removes the wind-up from character_sprite.gd (If we really want this effect, it should be implemented by actually moving the character instead of just the sprite, so physics can remain consistent)~~
Fixes the order of processing ground pound inputs. SIRIUS: This makes ground pounds work correctly on moving platforms.

# Issue(s)

Closes #179

Issue #244, opened 2023-09-17
Use a state machine for the player scene


GTcreyon:
**Describe the feature you'd like**
Rewrite player.gd and the player scene in general to make use of a state machine instead of a single monolithic script.

**Do you intend to contribute this yourself? If so, how do you plan on doing it?**
I can provide a state machine implementation designed with platforming in mind, derivative of https://github.com/Charpurrr/SuperMarioSolarEngine. I may be able to code the states themselves, but perhaps not.

Koopa1018:
Would facilitate solving of #158 as well.

PR #247, opened 2023-09-18
Apply pound windup animation to actual player position

### LINKED ISSUE #177: Unable to reach semi-solid platforms with a ground pound
**Describe the bug**
Looks like Mario can reach the semi-solid platforms with a ground pound, but it's just only an animation, so it doesn't actually increases the height.

**To Reproduce**
Try barely reaching the semi-solid platforms with a ground pound as you can see in the video here:

https://github.com/GTcreyon/SM63Redux/assets/13905392/5a0755c6-89c1-4187-a2c5-57f534ae34dd

**Expected behavior**
Mario/player should land on any semi-solid platforms with a ground pound to gain a little bit more height.

**System Information**
 - OS/Browser: Windows 10 / Avast Secure Browser (version 113.0.21244.127 (Official Build) (64-bit))
 - Game version: v0.1.6.alpha

Koopa1018:
# Description of changes
The pound spin animation appears to lift the player up slightly before the drop. This PR makes it so that this lifting isn't just visual, but rather actually affects the player's position.

This does have the side effect of giving the player a new movement option. Players likely assumed they already had this option, so this probably isn't a huge change to the play experience. Nonetheless, it deserves a mention.

Shout out to @Kuma-Boo for coming up with the apply-to-velocity method in PR #241. It's simple enough that I can't believe I ever considered applying the lift directly to the player position.

# Issue(s)
Closes #177.

### DIFF
--- a/classes/player/character_sprite.gd
+++ b/classes/player/character_sprite.gd
@@ -10,8 +10,6 @@ const SLOW_SPIN_END_SPEED = 0.40
 const SLOW_SPIN_TIME = 60
 
 const POUND_ORIGIN_OFFSET = Vector2(-2,-3) # Sprite origin is set to this during pound spin
-const POUND_SPIN_RISE = 1 # How much the player rises each frame of pound
-const POUND_SPIN_RISE_TIME = 15
 
 const SLIDE_MAX_SPEED: float = 5.0
 
@@ -180,11 +178,6 @@ func _physics_process(_delta):
 						
 						# Offset origin's X less at the start of the spin. (Looks better!?)
 						position.x *= parent.pound_spin_factor
-						
-						# A little rising as we wind up makes it look real nice.
-						position.y = POUND_ORIGIN_OFFSET.y
-						position.y -= POUND_SPIN_RISE * min(parent.pound_spin_frames,
-							POUND_SPIN_RISE_TIME)
 				parent.S.SPIN:
 					spin_logic()
 				parent.S.CROUCH:
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -921,7 +921,7 @@ func player_fall() -> void:
 		vel = Vector2.ZERO
 		# A little rising during the wind-up makes it look real nice.
 		if pound_spin_frames <= POUND_SPIN_RISE_TIME:
-			vel.y -= POUND_SPIN_RISE_AMOUNT
+			vel.y = -POUND_SPIN_RISE_AMOUNT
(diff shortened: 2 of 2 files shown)


classes/player/character_sprite.gd
```diff
@@ -180,11 +178,6 @@ func _physics_process(_delta):
 						
 						# Offset origin's X less at the start of the spin. (Looks better!?)
 						position.x *= parent.pound_spin_factor
```
Kuma-Boo, line 180: This line seems to cause a slight animation hiccup when the spin ends.

classes/player/player.gd
```diff
@@ -913,7 +917,11 @@ func ground_failsafe_condition() -> bool:
 func player_fall() -> void:
 	var fall_adjust = vel.y # used to adjust downward acceleration to account for framerate difference
 	if state == S.POUND and pound_state == Pound.SPIN:
+		# Don't move during the pound spin.
 		vel = Vector2.ZERO
+		# A little rising during the wind-up makes it look real nice.
+		if pound_spin_frames <= POUND_SPIN_RISE_TIME:
+			vel.y -= POUND_SPIN_RISE_AMOUNT
```
Kuma-Boo, line 924: Since vel is set to Vector2.ZERO a few lines before, this can probably just be `vel.y =` instead of `vel.y -=`

Kuma-Boo (changes requested):
Looks mostly good! One thing I noticed is that since we're moving the actual character now, the camera tracks the player's position a little too well, altering the feel of the move slightly.

It's probably something that can be tackled in a different PR, but it's something to keep in mind.

classes/player/player.gd
Koopa1018: While you're not wrong, that's an incredible nitpick. If it's not hurting readability, I'm inclined to just leave that.

classes/player/character_sprite.gd
Koopa1018: Confirmed, but probably not related. Was planning to fix in a separate PR.

Koopa1018:
Ooh, good thought. I hadn't considered camera feel.

classes/player/player.gd
Koopa1018: You know what, yeah, your suggestion is easier to understand.

GTcreyon (approved):
All looks good, although this continues to add complexity to the player script. See PR #244 for a potential solution. Happy to merge.

PR #256, opened 2023-09-22
Disable some toolscripts

### LINKED ISSUE #255: Paratroopas make version control noise
**Describe the bug**
Since the 4.1 update, Paratroopas' per-scene materials generate a new UID every time a level is opened containing them. These UID changes show up in version control, but are also completely useless and distract from actually-important changes being made.

**To Reproduce**
1. In the Godot editor, open any level with Paratroopas in it.
2. Save the level, without even making changes.
5. View the level in your Git client of choice.

**Expected behavior**
Level files only appear in version control when intentional changes were made.

**Screenshots**
image

**System Information**
 - OS/Browser: Windows 10
 - Game version: 4.1 branch
 - Other specs: VCS is GitHub Desktop

**Additional context**
Possibly solve in similar way to #227?

GTcreyon:
# Description of changes
Disable some toolscripts that are no longer relevant, since the scene editor will not be used for future level creation.

This is on the 4.1 branch, because the toolscript syntax changed in 4.1.

# Issue(s)
I hoped this would resolve #255, but it doesn't.

### DIFF
--- a/classes/console/console.gd
+++ b/classes/console/console.gd
@@ -270,7 +270,7 @@ func _input(event):
 	logger.offset_top = -40 - (Singleton.line_count + 1) * (logger.get_theme_font("normal_font").get_height() + 1)
 
 func display_completion(query: String, options: Array):
-	suggestions_log.text = ""
+	suggestions_log.clear()
 	
 	var sorted_options = []
 	for option in options:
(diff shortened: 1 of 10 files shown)


Koopa1018:
I wouldn't consider this part of the 4.1 project. It is wise to target 4.1 since that's soon going to replace `master`, but it doesn't actually fix problems _caused_ by the upgrade, so I'm going to untag this.

GTcreyon:
Understood.

Kuma-Boo (approved):
Looks good! Looking forward to when work on the level designer can begin.

Issue #207, opened 2023-07-29
Touch buttons sometimes fill the screen by default


Koopa1018:
**Describe the bug**
Multiple feedback tickets (not all, but multiple) have complained about unplayably large touch buttons on Android. From their phrasing, it seems this may have been the default button size for these people. (Additionally, today a person on Discord explicitly requested the addition of touch button resizing--which suggests people are not aware of the existing option to resize the controls.)

The exact sizes being complained about appear to vary slightly, but most recognizable complaints can only fit 3 buttons vertically on the screen (which leaves little horizontal room--this is more a problem in One-Finger control style than in New). I have seen one chain of 3 feedback posts (the descriptions of which are just "D") that had buttons the size of the screen; it's not clear whether this was 

**To Reproduce**
Unknown. Probably depends on users' screen resolutions, may also involve chosen touch control layouts.

**Expected behavior**
Regardless of screen size or any other factor, touch controls never start out too big to comfortably play the game around.

**Screenshots**
From ticket &e41ad. Screen resolution: 2220 x 1080. Description: "The game is Cooper the buttons are too big and blocking the screen shouldn't be out of it's finally"
![image](https://cdn.discordapp.com/attachments/937764234312450068/1128942515135979520/screenshot.png)
From ticket &86fe7. Screen resolution: 1920 x 1200. Description: "I'm on an android galaxy Tab A and the buttons are too big"
![image](https://cdn.discordapp.com/attachments/937764234312450068/1133439764650020984/screenshot.png)
From ticket &b5dd7. Screen resolution: 2000 x 1200. Description: "D"
![image](https://cdn.discordapp.com/attachments/937764234312450068/1132378884084613192/screenshot.png)

**System Information**
 - OS/Browser: Android
 - Game version: v0.1.6.alpha
 - Other specs: varies per report

PR #261, opened 2023-09-22
Fix back button scaling

GTcreyon:
# Description of changes
Fix an issue where the tappable back buttons on some menus scale poorly

### DIFF
--- a/gui/pause/options/touch_menu.tscn
+++ b/gui/pause/options/touch_menu.tscn
@@ -69,6 +69,7 @@ layout_mode = 2
 size_flags_horizontal = 6
 texture_normal = SubResource("1")
 texture_pressed = SubResource("2")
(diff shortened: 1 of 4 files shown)


Koopa1018:
This problem wasn't caused by the 4.1 update, so I'm going to untag for that.

Koopa1018:
Oh yes--sounds like this also fixes #207.

GTcreyon:
I don't think this was caused by 4.1.
Here's how it looks without the fixes:

Koopa1018:
Ahhh...giant in the pause menu, tiny in the options menu. Gotcha.

AidenBeresford (approved):
Looks good to me, I don't see how these changes could have any drawbacks. I'm going to approve, but it might still be good for someone else to review as well. Up to you.

PR #267, opened 2023-09-23
Un-ignore export_presets.cfg

GTcreyon:
# Description of changes
Remove export_presets.cfg from the .gitignore, so that changes can be tracked properly. This is now possible in Godot 4.1, since export paths are now relative, keystores are managed via editor settings, and sensitive data now goes in export_credentials.cfg.

(This commit does not claim that the export_presets.cfg file being created here is necessarily perfect. Consider checking over it for mistakes and misconfigurations.)


export_presets.cfg
```diff
@@ -0,0 +1,444 @@
...
+architectures/arm64-v8a=true
+architectures/x86=false
+architectures/x86_64=false
+version/code=7
+version/name="0.1.6"
+package/unique_name="com.sm63.redux"
+package/name="Super Mario 63 Redux"
+package/signed=true
+package/app_category=2
+package/retain_data_on_uninstall=true
```
Koopa1018, line 35: Out of curiosity, what data is there to retain on uninstall? Is it just settings, or is this looking ahead to user-created levels?

Koopa1018 (approved):
I've looked over everything, and it all seems in order.

Kind of annoying that the export path is also versioned...them's the breaks, I guess ¯\_(ツ)_/¯

export_presets.cfg
GTcreyon: Settings, and yes, user levels.

PR #272, opened 2023-09-24
Enhanced auto completion of commands

### LINKED ISSUE #263: Console commands don't go through on first confirm
**Describe the bug**
Since auto-completion was added to the console, typing commands straight through has become several steps less convenient.

In released alpha 0.1.6, I can type e.g. `fludd hover` and press enter, and the command immediately takes effect.

In the repo, ever since #189 was merged, I type the command and press enter, and rather than taking effect, an extra space is added to the end of the command. Once the space is added, pressing enter again makes the command go through as expected.

This does not occur with commands which end in a number argument, such as `health 8` and `w t1 4`.

**To Reproduce**
1. Open the debug console.
2. Type any valid command whose final argument is not a number.
3. Press enter to submit the command.
4. Press enter to submit the command again.

**Expected behavior**
Typed commands go through instantly, without having to add an extra keystroke.

**Screenshots**
N/A

**System Information**
 - OS/Browser: Windows 10
 - Game version: 4.1 branch (not tested in `master`, may also occur there)
 - Other specs:

**Additional context**
N/A

siddharthroy12:
# Description of changes
Command will get executed directly instead of adding a space if the input box text is same as auto completion text

# Issue(s)
Closes #263


jaschutte:
Could you resolve the merge conflict?

siddharthroy12:
@jaschutte Done

jaschutte (approved):

PR #273, opened 2023-09-24
Fixed moving platform pushing player if not on platform

### LINKED ISSUE #178: Moving semi-solid platforms pushes player
**Describe the bug**
Moving semi-solid platforms push Mario/player, due to collision problem.

**Video on how to reproduce**

https://github.com/GTcreyon/SM63Redux/assets/13905392/45758efc-14e1-46f6-843f-342f10c78570

**Expected behavior**
Player shouldn't get pushed, so it will only stand on moving platforms to carry around.

**System Information**
 - OS/Browser: Windows 10 / Avast Secure Browser (version 113.0.21244.127 (Official Build) (64-bit))
 - Game version: v0.1.6.alpha

**Additional context**
This could also affect birds pushing the player, but I'm not sure if it's a feature or not.

siddharthroy12:
# Description of changes
Fixed moving platform pushing player even when it's not on the platform.

# Issue(s)
Closes #178


GTcreyon (changes requested):
This does help, but it doesn't quite solve the problem. Mario's origin point is not at his feet, it's in the center of his body, so the platform can still carry him if it's a more shallow intersection. Moreover, this origin point is inconsistent for different entities. Perhaps this could be resolved by using two Area2Ds - one for the initial check, and another to check if the entity is actually underneath the platform?

image

Like so. The lower one cancels out the upper one.

siddharthroy12:
I like the two Area2D idea. I think that will work for all cases

siddharthroy12:
I have implemented this but the lower box is one pixel lower to prevent the edge collsion.
image

jaschutte (approved):
Seems to work to me!

jaschutte:
@GTcreyon could you double check, as you requested changes

GTcreyon:
> @GTcreyon could you double check, as you requested changes

I'm not too pressed about it, as long as it works.

jaschutte:
Yeah but I can't merge it until you approved the changes x3

GTcreyon:
Oops.

GTcreyon (approved):
I'm sure it's fine. I'm tired.

PR #275, opened 2023-09-27
Refactor pound spin

Koopa1018:
# Description of changes
Moves the pound spin animation from `player.gd` to `character_sprite.gd`. The effect is purely visual (can't affect FLUDD angle since you can't FLUDD during a pound), so it probably shouldn't be lumped in with the player physics.

Not to be confused with the pound spin _rise_ animation, which _does_ affect player physics. See #247 for that.

# Issue(s)
Makes good on a longstanding `TODO` in the code.

# Before merge:
- [ ] There's an annoying pop forward on the X axis when the spin ends (also brought up by Kuma in #241). I want to fix that, but I'm still not 100% sure how to go about it. (The offending line has at least been identified.)


GTcreyon:
I suspect this will be obsoleted by #244. Should this be closed?

Koopa1018:
Hmm, yeah, good call

PR #278, opened 2023-09-29
Move player SFX bank to a resource

### LINKED ISSUE #108: Move player SFX bank to a resource
Currently, the SFX bank in the player script is at the top, as an obscenely long constant dictionary. I feel that we could do this more efficiently if we used a custom resource class.

I may implement it later, but not right now.

AidenBeresford:
# Description of changes
Adds a new SFXBank resource which uses export variables to better organize (and possibly optimize?) the current sound effect system, a hard-coded dictionary in the player script.

# Issue(s)

Closes #108


GTcreyon:
How's this going? Some input from @Koopa1018 might be nice.

PR #282, opened 2023-10-29
Fix t1r4 color block being white instead of color

GTcreyon:
# Description of changes
In Tutorial 1 Room 4, the first rotating block is white translucent. It should be color. This fixes that.


Koopa1018 (approved):
Huh! Has this been wrong since the beginning? Because I had totally forgotten that this block was colorful in OG....

PR #285, opened 2023-11-21
Allow turning around while crouched

### LINKED ISSUE #284: Not being able to turn around while crouched feels sluggish
Excuse the title, I couldn't come up with something better.

I propose to allow the player to turn around while crouched. This would allow the player to correct the direction in which they are facing while crouched, instead of having to stand up first before turning around; similar to how it works in many modern Mario games.

This also makes it more comfortable for the player to crouch to perform a backflip, as they will no longer need to uncrouch to turn around in case they suddenly decide they need to do a backflip the other way, which risks an accidental dive. This scenario is the one that I believe would benefit the most from this proposal.

Yes, this is something quite easy to implement and I already have an idea how to do it. Adding the crouching state as a valid state to allow a turn around will do the job.

**Additional context**
I made this demonstration video to show how the proposal would work in game:

https://github.com/GTcreyon/SM63Redux/assets/75398915/72529333-b00e-499a-a07e-e8dcf65e8ae1

And this video shows the current behavior when trying to turn around while crouching:

https://github.com/GTcreyon/SM63Redux/assets/75398915/3f2d5d86-e353-4a39-9b1c-5a822ee875e1

Kawtious:
# Description of changes
Allow turning around while crouching by adding the crouching state to the list of valid cases that allow turning around.

Quoting myself from #284:
> This would allow the player to correct the direction in which they are facing while crouched, instead of having to stand up first before turning around; similar to how it works in many modern Mario games.
>
> This also makes it more comfortable for the player to crouch to perform a backflip, as they will no longer need to uncrouch to turn around in case they suddenly decide they need to do a backflip the other way, which risks an accidental dive. This scenario is the one that I believe would benefit the most from this proposal.

However, even though this is but a tiny change in code, I do think that this could be a slightly major change that will impact movement tech; and I think it'd be appropriate to discuss about it before making these changes... unless I'm overthinking it. I'd appreciate some feedback regardless.

Let me know about any suggestions, and if there should be any balancing that should be done beforehand.

# Issue(s)
Closes #284

### DIFF
--- a/classes/player/player.gd
+++ b/classes/player/player.gd
@@ -685,15 +685,27 @@ func player_control_x() -> void:
 	var dir = get_walk_direction()
 	if dir != 0:
 		if pound_state != Pound.SPIN:
-			var can_turn_around: bool = state & (
-					S.NEUTRAL
-					| S.SPIN
-					| S.TRIPLE_JUMP
-					| S.ROLLOUT
-					| S.CROUCH
-				) or crouch_resetting
-			if can_turn_around:
-				facing_direction = dir # Will never be 0
+			# Check if the player can turn around while crouched.
+			# This only needs to be applied when trying to face the opposite direction.
+			if state == S.CROUCH and facing_direction != dir:
+				# We apply a delay between the input and the action to give the player a window
+				# to perform a backflip with maximum force; otherwise, the player would just
+				# instantly turn around and the backflip would come out all messed up.
+				# The player would require frame-perfect precision to perform such backflip, if not
+				# for this delay.
+				manage_crouch_turn_around_delay()
+				if can_turn_around_while_crouched:
+					facing_direction = dir
+					reset_crouch_turn_around_delay()
+			else:
+				var can_turn_around: bool = state & (
+						S.NEUTRAL
+						| S.SPIN
+						| S.TRIPLE_JUMP
+						| S.ROLLOUT
+					) or crouch_resetting
+				if can_turn_around:
+					facing_direction = dir # Will never be 0
 			if state != S.CROUCH or crouch_resetting:
 				if (grounded or (state & (S.ROLLOUT | S.BACKFLIP | S.DIVE | S.NEUTRAL) and ground_except)) and !swimming:
 					if state == S.POUND:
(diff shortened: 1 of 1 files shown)


GTcreyon:
Hi! Sorry for taking a while to get to this. Remember that you can request a review from members of the team using this section:
image
It gives you some suggestions of who you could request a review from, if you don't know who would be appropriate.

Regarding the change itself, I'm conflicted. While I agree this makes the movement more responsive, it's worth considering that the purpose of a backflip is to send you *backwards.* That means you're likely to be holding the direction opposite to the one you're facing. Consider this case:

https://github.com/GTcreyon/SM63Redux/assets/67928545/63dfa0e9-a37f-4bd3-b5d1-cda28dc2b0b5

I want to backflip backwards. With this new change, I have to hold back *after* I backflip. If I do it before, I get turned around, and my backflip is all messed up.

I'd like to hear what others have to say. My vote is currently on "no", but I'm not super convicted on it.

Kawtious:
> ### **Adding a short delay before being able to turn around after crouching**
> 
> The delay should be just long enough to let the player comfortably, but quickly, perform a long backflip. After the delay, the player would be able to turn around while crouched. Turning around _while_ crouched could re-apply the delay, in case the player wants to turn around and then do a long backflip.

This sacrifices a bit of the responsiveness that the previous changes provide, but I think this is a good middle ground. There is now a short (although decently sized) window where the player can input a long backflip without instantly turning around.

https://github.com/GTcreyon/SM63Redux/assets/75398915/7549011b-d1dd-4a4b-8da2-f1bee0244306

Here's a flowchart to complement the implementation explanation I made in the comment above:

SM63R Crouch Turn Around Flow Chart

jaschutte:
I like your proposed solution, I think it's a great quality of life addition.

The issue is indeed the player re-write, but I see no harm in merging this before that time. Even if it were for a reference implementation. Thanks for the flowchart by the way!

Opinions @GTcreyon @Koopa1018 ?

GTcreyon:
> I like your proposed solution, I think it's a great quality of life addition.
> 
> The issue is indeed the player re-write, but I see no harm in merging this before that time. Even if it were for a reference implementation. Thanks for the flowchart by the way!
> 
> Opinions @GTcreyon @Koopa1018 ?

I don't think there's much of a reason to merge this as-is. I think the PR should be closed, personally.

However, I do like the alternative implementation, and the flowchart was actually very helpful. I think I'll close this PR, but I'll look into implementing it when doing the rewrite.

PR #286, opened 2023-11-25
Fix Level Designer bugs

Koopa1018:
# Description of changes
Fixes most of the bugs I could find in the Level Designer after the 4.1 merge. Since the LD didn't run in build 0.1.6.alpha, I had no idea what was from 4.1 or not--so I just fixed everything I ran across.

Changes include:
- Comment, typehint, and name-clarify everything I touch, as usual.
- UI refactors to get everything working again, and hopefully more future-proof too.
- Several LD-exclusive scripts were moved into the LD script folder.
- Item panel's scroll area was changed to use a Godot native ScrollArea node. My eventual plan is to switch many LD widgets to native Godot nodes, but making the switch for this one actually fixed bugs--so I did it now instead of waiting for next PR. Note that because formatting graphics for the new system is outside the scope of this PR, the new scrollbar still uses Godot's built-in theme.
- Now all items can be added without crashing on play.
- Rotating blocks unexpectedly now show up in the items list. (They have no icon and can't be placed, but they do show up!)
- The "Open Level" dialog now acts like an open dialog instead of a save dialog. Before, selecting a level would confirm overwrite before it could open it!
- The glow shader was substantially rewritten to avoid deforming at high outline thicknesses. This change has had knock-on effects throughout the project; only _most_ of the bugs introduced have been fixed thus far.
- Fixed a bug where the LD music would jump to near-silent at the start of fading out.

# To do before review:
- [x] ~~As mentioned earlier, one bug remains in the glow shader...the pause menu's shine icon has an outline which extends too far on two sides.~~ Going to deal with this separately, as #289.
- [x] In addition, item selection still needs to _account_ for the new glow shader. Currently, selecting items from multiple atlases will cause some selected objects to bug out or even become invisible.
- [x] Is Hermes meant to animate while moving forwards? (EDIT: Shima says no, only when turning. This is the current behavior, so we're good.)
- [x] I've been keeping my to-do list in a text file at the project's root. Best to delete that before we merge.

# Out of scope for this PR:
- The serializer's condition has improved, but loading is still broken in multiple cases. We agreed on Discord that it would be better to build a new, more maintainable serializer than to fix the current one. That's a project unto itself.
- The item pane scrollbar still uses the Godot default theme. Graphics changes are out of scope for this PR, and scrollbars in particular will require custom draw code to look the same as they did before.

# Issue(s)
No GitHub issues, but these bugfixes are needed for LD work to begin in earnest.


Koopa1018:
@GTcreyon @jaschutte poke

GTcreyon:
> poke

Sorry for the delay. I'll get on this soon.

classes/solid/log/log.gd
```diff
@@ -5,4 +5,4 @@ extends StaticBody2D
 
 func set_disabled(val):
 	disabled = val
-	set_collision_layer_value(0, 0 if val else 1)
+	set_collision_layer_value(1, 0 if val else 1)
```
GTcreyon, line 8: Check #291 for conflicts.

classes/solid/telescoping/telescoping.gd
```diff
@@ -56,9 +56,9 @@ func set_disabled(val):
 		safety_net = $SafetyNet
 	safety_net.monitoring = !val
 	
-	var carrier_node = get_node(get_path())
```
GTcreyon, line 59: It was like this for a reason. I don't recall why. Either way, this can be made obsolete by standardizing the way the `disabled` flag works, so that we don't need to do this check.

scenes/menus/level_designer/follow_viewport_workaround.gd
```diff
@@ -2,8 +2,10 @@ extends CanvasLayer
 
 # As of godot 3.5, buttons inside of a canvaslayer with follow viewport enabled do not work
 # So this is basically a script which makes the canvaslayer follow the viewport with it still working
+# TODO: Game is on 4.1 now. Check if this is still true.
```
GTcreyon, line 5: No longer applies. This script can be ditched, and Follow Viewport can be safely enabled.

scenes/menus/level_designer/ldui/ld_camera.gd
```diff
@@ -42,12 +44,14 @@ func _process(_delta):
 	if i_right and Input.is_action_pressed("debug"):
 		LD.save_code()
 
-func object_load(id):
-	if !objects_loaded.has(id):
-		objects_loaded[id] = load("res://classes/items/" + str(id) + "/" + str(id) + ".tscn")
-	return objects_loaded[id]
 
 # Disabled this for now as I planned to use left drag for selecting items
```
GTcreyon, line 48: This can go. The planned feature has been implemented.

GTcreyon (changes requested):
Some minor changes requested. Hopefully now that the bulk of the issues are resolved, future PRs will be smaller and easier to review... Really good work, though.

classes/solid/log/log.gd
Koopa1018: Oh, good catch.

classes/solid/telescoping/telescoping.gd
Koopa1018: Same method as #291 seems like a good choice. 

If we're talking standardizing, I'd like to discuss making disabling a bit more granular--so the player always knows they're checking "disable collision" or "disable movement" or "disable interaction" rather than wondering, "does disabling a pipe make it non-solid, or just non-enterable?"

classes/entity/entity.gd
```diff
@@ -101,7 +101,7 @@ func _on_WaterCheck_area_exited(_area):
 func set_disabled(val):
 	disabled = val
 	_preempt_all_node_readies()
-	set_collision_layer_value(0, 0 if val else 1)
+	set_collision_layer_value(3, 0 if val else 1)
```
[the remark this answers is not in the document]
Koopa1018: Not sure I understand what you mean.... You want us to skip setting the collision layer?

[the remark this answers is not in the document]
GTcreyon: Sorry, I wasn't clear. I meant setting `disabled = val` on the *collision shape*, not the root node. Your method of disabling processing (like in the solid class) might make more sense, though.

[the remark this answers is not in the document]
Koopa1018: ahhh, ok. Ties back to standardizing the method, then.

Koopa1018:
Arright, that's all discussions resolved, merge issues resolved, and an extra bug caused by the merge fixed. Ready for merge now?

GTcreyon (approved):
Well-overdue for a merge. Good work!

PR #291, opened 2024-02-23
Generalize solid-object scripts

Koopa1018:
# Description of changes
Renames `log.gd` to `solid_object.gd`, and `big_rock.gd` to `solid_object_mirrorable.gd`. Also moves both of these scripts directly into `classes/solid` out of their subfolders. Neither of these scripts actually has anything specific to rocks or logs--one lets the LD disable a StaticBody2D, the other lets the LD disable _and mirror_ a StaticBody2D. This seems useful for many different solid object types.

# Issue(s)
Minor issue noted during #286.


Koopa1018:
Notes from Discord:
- Forgot to make collision polygons mirror with sprites. Mirrored rocks thus have wrong collision.
- Disabling is bugged, and would be better done via disabling anyway
- Hardcoded `$Sprite2D` node path is not ideal, should be an export variable

classes/solid/solid_object.gd
```diff
@@ -0,0 +1,22 @@
...
+
+@export var disabled = false: set = set_disabled
+
+
+func set_disabled(val):
+	disabled = val
+	
+	# Set the object's process mode as appropriate.
+	# When disable_mode = DISABLE_MODE_REMOVE, this makes the object
+	# non-solid when disabled.
```
GTcreyon, line 18: Very nice approach.

classes/solid/solid_object_mirrorable.gd
```diff
@@ -0,0 +1,90 @@
...
+			child.polygon = flip_points(child.polygon)
+		
+		# Flip collision shapes that can flip.
+		elif child is CollisionShape2D:
+			var shape := (child as CollisionShape2D).shape
+			if changed and mirror:
+				# On becoming mirrored, make a copy of the shape resource
+				# so we can mirror its data without messing up any other users
+				# of the shape resource (yes, the shapes are resources!).
+				shape = shape.duplicate() # Make and store the duplicate...
```
GTcreyon, line 53: I assume Godot will garbage-collect shapes that are no longer in use, to avoid a memory leak?

GTcreyon (approved):
Some notes, but nothing blocking.

classes/solid/solid_object_mirrorable.gd
Koopa1018: Yeah, Resource inherits RefCounted. I actually put a whole paragraph about that below the assignments...maybe I should've put it above and accepted the daunting-ness of it :/

GTcreyon: Oh, guh, I see that now. Thanks.

I think it's fine where it is, I just tuned it out because GitHub's web UI wraps the text in a weird way that makes it hard for me to pay attention to it.

PR #297, opened 2024-05-29
Make tilesets into a Resource

Koopa1018:
# Description of changes
This PR changes the terrain system to use a `TerrainSkin` resource instead of raw tileset template images.

This change has many benefits. Most notably, terrains can now use textures of arbitrary size, since the texture size is no longer hardcoded to match the specific template layout. Texture reuse between terrain lumps is another benefit, as well as the theoretical ability to add arbitrary data to terrain skins (e.g. footstep sound).

Terrain lumps now automatically update when the terrain skin is modified. Oddly enough, it turns out the tools needed to implement this feature technically existed in the old system...nothing like a total rewrite to help you see what's already there, I guess :/

To preserve the single-image convenience of the old system (and compatibility with all the assets we made for it), an import plugin has been written which allows importing images as terrain skins. It comes with features to modify which rects are used for creating textures, and options to either pack the sliced textures into the resource or store them as separate image files. As this is a vanilla Godot importer plugin, the file conversion is almost completely seamless, allowing images to be dragged directly onto `TerrainSkin` fields in the Inspector, and to still be edited directly in image editing programs.

# Issue(s)
Finishes much of [Taiga \#69](https://tree.taiga.io/project/gtcreyon-super-mario-63-redux/us/69) (ncei).


jaschutte:
With the dependency gone, is it ready for merge?

PR #299, opened 2024-06-02
Feature ld polygon editor

jaschutte:
Implements the polygon editor for the level designer. This PR aims to fully implement the polygon editor. This editor is supposed to work for all "things" which require a polygon to edit. 

## Todo list

- [x] Select polygons
- [x] Place polygons
- [x] Edit polygon vertices
- [x] Move polygon vertices
- [x] Erase polygon vertices
- [ ] Active edge selector (for water, is this edge have ripples? for terrain, should this edge have grass?)
- [ ] Proper UI/UX


Koopa1018:
With the current state, we have a polygon editor which is basically usable. Edge selection is going to be a whole new job's worth of work, and isn't essential for the immediate term. What say we get this merged as-is and handle the other two checkboxes later?

PR #300, opened 2024-06-08
Drastically optimize pause menu elements

Koopa1018:
# Description of changes
On profiling the normal gameplay loop, it was discovered that several pause menu elements had startlingly high performance costs. This PR optimizes them down to reasonable levels.

The bigger offender was the function `_get_joypad_buttons` in `rebind_option.gd`, which was profiled as using 11(!) ms of CPU time in each frame. This function returns an array of arrays of button names, indexed by logical button index and by controller brand. On inspection, it was found that this function was called once for every binding of every action, every frame, and doing the same 25 string translations every time (which suggests 25 disk reads). Assuming one binding for every action--which I know is lowballing it--that's `25 * 25 = 625` total translations per frame, literally 98% of which are repeated work.

Optimizing the script thus becomes a matter of reusing work. This can be done:
- within `_get_joypad_buttons`. Because some buttons share names between brands, some translated strings can be cached before constructing the return array, which skips 15 extra translations per function call. This step alone reduced CPU time from 11ms to 4ms.
- between actions and bindings. The result of calling `_get_joypad_buttons` doesn't depend on which action or mapping it's used for--only the selected locale. By storing the result in a static variable, it can be reused between instances, and the disk reads(?) involved in translation lookup can be skipped for all but the first user.
- between frames. Locale really only changes when the user sets it in the options menu, so for 99% of frames, `_get_joypad_buttons` returns the same result as last frame. `rebind_option` already has machinery to receive the engine's locale-change notification, so it's fairly trivial to only set `_get_joypad_buttons` when the game starts or the locale changes. Combined with the previous step, this reduced CPU time from 4ms to ~0.3ms, about level with `Entity._physics_process` and `DejitterGroup._physics_process`.
- Actually, the majority of `rebind_option`'s state doesn't change between frames. By adding (and connecting to) a signal to the reset-mappings button, it becomes possible to completely remove `_process` and instead update rebind options reactively. Naturally, this means `rebind_option` is no longer charting at all on most profiler frames.

Some similar string-caching-via-statics work was also done to `get_joypad_motion_name`. This is mostly for methodological consistency, but if somebody were to make very strange mapping choices ("wassup YouTube, today I'm gonna be playing SM63R using just the left and right analog stick!"), I could see it saving actual performance.

I also chose to optimize `level_info.gd`. This was much less of a problem script, but it still seemed appropriate to switch it from detecting locale via `_process` to detecting via engine notification, especially since `rebind_option` had shown me how to do that. This saves about 0.15ms--not much, but still seemed excessive for a static menu element.

Some mild refactors have also been sprinkled in, including obligatory reordering functions to standard and commenting everything I touch.

### DIFF
--- a/gui/hud/level_info.gd
+++ b/gui/hud/level_info.gd
@@ -9,11 +9,17 @@ var perm_scale = 1
 @onready var level_name = $Divider/MainContainer/LevelNamePanel/ScaleContainer/LevelName
 
 
-func _process(_delta):
-	# Update the font used, in case locale's changed.
-	if TranslationServer.get_locale().substr(0, 2) == "en":
-		level_name.add_theme_font_override("font", RUBY)
-		level_name.uppercase = true
-	else:
-		level_name.add_theme_font_override("font", BYLIGHT)
-		level_name.uppercase = false
+func _notification(what):
+	if what == NOTIFICATION_TRANSLATION_CHANGED:
+		match TranslationServer.get_locale().substr(0, 2):
+			"en":
+				# Fancy display font has all needed characters.
+				# Use it for the level name.
+				level_name.add_theme_font_override("font", RUBY)
+				# Display font has only uppercase glyphs.
+				level_name.uppercase = true
+			_:
+				# Display font may not have all needed characters.
+				# Use the standard font.
+				level_name.add_theme_font_override("font", BYLIGHT)
+				level_name.uppercase = false
--- a/gui/hud/level_info.gd
+++ b/gui/hud/level_info.gd
@@ -3,9 +3,14 @@ extends Control
 const BYLIGHT = preload("res://fonts/bylight/bylight.otf")
 const RUBY = preload("res://fonts/red/gui_red.fnt")
 
-@onready var level_name = $Divider/MainContainer/LevelNamePanel/ScaleContainer/LevelName
 
 var perm_scale = 1
+
+@onready var level_name = $Divider/MainContainer/LevelNamePanel/ScaleContainer/LevelName
+
+
+func _process(_delta):
+	refresh_caps()
 
 
 func refresh_caps():
@@ -15,7 +20,3 @@ func refresh_caps():
 	else:
 		level_name.add_theme_font_override("font", BYLIGHT)
 		level_name.uppercase = false
-
-
-func _process(_delta):
-	refresh_caps()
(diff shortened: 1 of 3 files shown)


gui/pause/options/rebind_option.gd
```diff
@@ -24,31 +65,38 @@ func _input(event):
...
 			# TODO: make these translatable
-			output += action.as_text()
+			output += bind.as_text()
+		# Add separator between this and the next binding.
 		output += ", "
+	# Trim off the final separator for visual cleanliness.
 	output = output.trim_suffix(", ")
-	return output
+	
+	key_list.text = output
```
GTcreyon, line 99: The game crashes here for me. "Invalid assignment of property or key 'text' with value of type 'String' on a base object of type 'Nil'."

I believe this happens because `update_list()` is called before `@onready` can trigger, when the translation change notification is received. I'm not sure why this didn't occur while you were working on it, perhaps Godot updates, or perhaps something to do with my own language settings? (Apologies for the belated response to your PR.)

gui/hud/level_info.gd
```diff
@@ -3,19 +3,23 @@ extends Control
...
-		level_name.add_theme_font_override("font", BYLIGHT)
-		level_name.uppercase = false
+@onready var level_name = $Divider/MainContainer/LevelNamePanel/ScaleContainer/LevelName
 
 
-func _process(_delta):
-	refresh_caps()
+func _notification(what):
+	if what == NOTIFICATION_TRANSLATION_CHANGED:
+		match TranslationServer.get_locale().substr(0, 2):
```
GTcreyon, line 14: A similar issue occurs here, with a similar solution. (See my other comments.)
```suggestion
		if level_name == null:
			await ready
		match TranslationServer.get_locale().substr(0, 2):
```

gui/pause/options/rebind_option.gd
```diff
@@ -173,4 +245,9 @@ func _get_action_map() -> Dictionary:
 func _notification(what: int) -> void:
 	match what:
 		NOTIFICATION_TRANSLATION_CHANGED:
+			# Update cached strings (but only if the first instance hasn't
+			# already done that).
+			if locale_current != TranslationServer.get_locale().substr(0, 2):
+				_update_translations()
+			# Update the list.
 			update_list()
```
GTcreyon, line 253: The aforementioned crash can be resolved like so:
```suggestion
			if key_list == null:
				await ready
			update_list()
```

GTcreyon (changes requested):
Generally all good, just some crashes that need resolving.

PR #303, opened 2024-08-01
Fix background scaling in sized-up windows

Koopa1018:
# Description of changes
This has been on the backburner for way too long! Fixes background scaling so that sized-up windows don't show "backstage" parts of the background.

The parallax is not 100% consistent between scaling levels. I'm not sure how to fix this, but it's a very subtle problem that's (currently) easy to miss, so I'll leave it for now. We'll see how new backgrounds adapt to it....

# Issue(s)
Been reported many times, including by @GTcreyon.

### DIFF
--- a/backgrounds/zooming_parallax_background.gd
+++ b/backgrounds/zooming_parallax_background.gd
@@ -1,0 +1,7 @@
+extends ParallaxBackground
+
+
+func _process(delta):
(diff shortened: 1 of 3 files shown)


GTcreyon (approved):
All looks good. The detailed commenting is appreciated as always!

PR #304, opened 2024-08-17
Pipe offsetting when loading a level in the LD

cruglet:
# Description of changes
Offset the pipe's sprite and collision a few pixels downward

# Issue(s)
Pipe would be offset by a few pixels vertically when placed


GTcreyon (approved):
Seems good. (Remember to request review when you create a PR, though :P)

PR #305, opened 2024-08-29
Chinese localization update

WestCreeper:
Added chinese localization and font with Chinese characters.


fonts/bylight/BylightV24-8-11MYIL.otf
GTcreyon, line 1: This file needs to be renamed to "`bylight.otf`".

GTcreyon (changes requested):
The font file should be renamed to `bylight.otf`. Also, could you inform us of what font you used, so we can check copyright information and give credit to the designer?

GTcreyon (approved):
Thank you for your work!

PR #306, opened 2024-12-27
Refactor filesystem

cruglet:
# Description of changes
Most of the ingame content has been moved into the `game` folder with split directories for textures/scripts/sfx/etc. This provides an easier framework for the refactoring of underlying systems, as well as future implementations for other systems to come (and organization, of course). The level designer and pipescript have not really been touched, as those can be addressed later when refactored. The only change made to those would be the moving of pipescript's assets from godot into the respective pipescript folder, under `scene` (the rust source code is under `source`).

The only part of this refactor that I sort of have a issue with is the UI and "classes". I'd be lying if I didn't want to just tear most of that down and start anew, but I'm still unsure how important those elements are going to be in the long run, so they'll stay mostly untouched in this refactor.

Here is a summary of the new filesystem:

**android** - Untouched  (still am unsure if this should be moved somewhere)

**console** - The in-game console (/)

**core** - Basically whatever does not fit in `game` like the audio bus layout or specific shaders.

**game** - All assets [textures/scenes/scripts/sfx/music] for the game
- classes: Any former class are organized here, as opposed to being scattered throughout the filesystem
- locale: The folder for fonts & translations
- music: All of the game's music
- objects: Any entity/sprite with their textures/sfx/scripts/etc.
- ui: The UI for the game (HUD, Menus, Screens, etc.)
- world: Mostly for unsorted stuff like backgrounds, terrain, water, death & camera zones, etc. 

**level_designer** - Untouched, just moved out of its original place

**meta** - Any assets for publications of the game or for the README

**pipescript** - Contains everything for pipescript, including its respective Godot scene and resources.

# Issue(s)
A lot of things are broken right now. Good news is we were gonna rewrite those systems anyway, so its not that big of a deal in the long run 🤷


jaschutte:
Although I like the new organization, how much of the game is still, functional after all of this? A lot of code uses absolute pathing for resources, I imagine this could create quite the fuss with a lot of systems, such as the level editor.

Also, does moving the files under a sub-folder erase tracking? If so, it may cause issues with other branches active right now too.

I'll do a more thorough check tomorrow, my main point of contention is erasure of tracking, that would make it all old issues harder to track. But I'm overall supportive of the changes

cruglet:
> Although I like the new organization, how much of the game is still, functional after all of this? A lot of code uses absolute pathing for resources, I imagine this could create quite the fuss with a lot of systems, such as the level editor.

This is true, and I wouldn't be surprised if the vast majority of systems did not work correctly. The thing is, those exact same systems were due for a rewrite anyway, either because we have to remove specific elements of source code or the code itself is old/composed of a mixture of band-aid solutions and spaghetti code (like the former player script). Regardless of whether some systems break or not, the _direct_ process to fix those issues would be as easy as reassigning paths, simply following the engine's errors would be more than enough to fix those issues. Which also gives us the opportunity to revisit this code and see if any improvements/optimizations can be made.

> Also, does moving the files under a sub-folder erase tracking? If so, it may cause issues with other branches active right now too.

This was my major concern with the refactor. To my knowledge, tracking is not erased with `git mv`, however I'm unsure if moving files in Godot's filesystem compensates for that. Regardless, since we're kinda obligated to rewrite a lot of systems, it wouldn't be so bad if it did, anyway. Older issues can be addressed _during_ the refactoring of other systems, and the only active branches I'm aware of is the player refactor and the ld-polygon + ld-widget branches.. Those wouldn't be too hard to port over, especially since the ld was barely touched (for this reason) and I was gonna rewrite the player statemachine anyway 👍

cruglet:
Closed due to solar integration

